PR #30, opened 2019-09-10
update website footer

ymatsuda:


website/siteConfig.js
```diff
@@ -31,7 +31,7 @@ const siteConfig = {
 
   // Used for publishing and more
   projectName: 'test-site',
```
biplap-sarkar: Should we update this as well?

PR #34, opened 2019-09-19
fix potential deadlock

whynick1:
To explain the cause, we need to understand three methods.

1) `append()` enqueue append task to queue.
2) `doAppend` dequeue append task from queue.
3) `resolveAppendRequests()` commits transaction.
4) `append()`, `doAppend`  and `resolveAppendRequests()` share the lock.
5) when append queue is full, `append()` will `wait()` while calling `queue.enqueue()`, which is fine. Because normally, `doAppend()` will unblock `append()` by calling `queue.dequeue(BATCH_SIZE)`
6) however, there is a corner case when load is heavy. When `append()` was called many times in a short time and gets blocked due to queue full, it holds the lock so neither `resolveAppendRequests()` nor `doAppend()` could move forward.
7) finally, we have a deadlock

To fix it, we need to tell `append()` to give up on it lock when the queue is "almost full".

### DIFF
--- a/waltz-server/src/main/java/com/wepay/waltz/store/internal/StoreSessionImpl.java
+++ b/waltz-server/src/main/java/com/wepay/waltz/store/internal/StoreSessionImpl.java
@@ -172,6 +172,9 @@ public class StoreSessionImpl implements StoreSession {
                     } catch (InterruptedException ex) {
                         Thread.interrupted();
                     }
+                    if (!running) {
+                        return;
+                    }
                 }
 
                 if (requestQueue.enqueue(request)) {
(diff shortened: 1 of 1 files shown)


waltz-server/src/main/java/com/wepay/waltz/store/internal/StoreSessionImpl.java
```diff
@@ -165,6 +165,15 @@ public void append(StoreAppendRequest request) throws SessionClosedException, Ge
                     throw new GenerationMismatchException();
                 }
 
+                // Wait until the queue has enough space.
+                while (requestQueue.size() > BATCH_SIZE * 2) {
```
ymatsuda: Need to break out of the loop when `running` became false.

whynick1: Don!

whynick1: Done!

PR #40, opened 2019-09-25
added onApplication and onLockFailure to TransactionContext

ymatsuda:
This PR adds two new callback methods to `TransactionContext`
1. onApplication()
2. onLockFailure()

`onApplication()` is called when the transaction generated by the transaction context is applied to the client database through `WaltzClientCallbacks.applyTransaction()`.
`onLockFailure()` is called when an append request is rejected due to a lock failure. This doesn't change the automatic retry by Waltz client. `TransactionContext` will be re-executed later.

### DIFF
--- a/waltz-client/src/main/java/com/wepay/waltz/client/TransactionContext.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/TransactionContext.java
@@ -69,7 +69,12 @@ public abstract class TransactionContext {
     }
 
     /**
-     * A method that is call on failure of optimistic locking
+     * A method that is called on failure of optimistic locking.
+     * There can be a spurious lock failure with a very small probability.
+     * A spurious lock failure is the case that the lock has failed even though the locks high-water mark has not been changed.
+     * So, this callback should treated as an informational notification.
+     * Waltz client will automatically invoke the execute method again after the client consumed a transaction that
+     * caused the lock failure. And the execute method should decide whether the transaction should be retried or aborted.
      */
     public void onLockFailure() {
     }
--- a/waltz-client/src/main/java/com/wepay/waltz/client/TransactionContext.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/TransactionContext.java
@@ -53,12 +53,25 @@ public abstract class TransactionContext {
     }
 
     /**
+     * A method that is called on application of transaction through
+     * {@link WaltzClientCallbacks#applyTransaction(Transaction)}.
+     */
+    public void onApplication() {
+    }
+
+    /**
      * A method that is called on exception.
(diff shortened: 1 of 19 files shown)


waltz-client/src/main/java/com/wepay/waltz/client/TransactionContext.java
```diff
@@ -61,4 +68,10 @@ public void onCompletion(boolean result) {
     public void onException(Throwable ex) {
     }
 
+    /**
+     * A method that is call on failure of optimistic locking
```
abhipardhu: javadoc nit: "called"

waltz-client/src/main/java/com/wepay/waltz/client/TransactionContext.java
```diff
@@ -61,4 +68,10 @@ public void onCompletion(boolean result) {
...
+     */
+    public void onLockFailure() {
```
abhipardhu: Do you think it would be useful if application can decide whether or not to retry that log, probably by returning a boolean to this method?

ymatsuda: There can be a spurious lock failure (with a very small probability) due to a probabilistic nature of our lock table. I thought it is not safe to make retry decision in `onLockFailure()`.
Do you think it is useful to have an option to abort transaction even when it may be a false lock failure?

abhipardhu: Ah... I see, got it.

No, I think it is okay not to have that option.

(I was worried about usecases where synchronous api calls submit logs to Waltz n wait for those, but, in case of lock failure, app client can implement an expiry-like mechanism to return the api call rather than timeout waiting for the log to be retried n succeeded. When the api-call is retried, app client can check in its DB n determine whether or not to resubmit the log. If resubmitted, only one of the resubmitted or the being-retried logs will be successful due to optimistic locking.)

waltz-client/src/main/java/com/wepay/waltz/client/internal/TransactionMonitor.java
```diff
@@ -184,21 +185,45 @@ public TransactionFuture register(ReqId reqId, long registrationTimeout) {
...
+            }
+            return null;
+        }
+    }
+
     /**
      * If the transaction monitor is not closed,
      * the corresponding {@link TransactionFuture} of the {@code reqId} is completed with {@code true}.
      *
      * @param reqId the {@code ReqId} of the transaction that was committed to a Waltz server.
```
criccomini: Nit: add @return so javadocs doesn't complain

criccomini:
Actually, on second glance, I think compatibility is OK...

PR #46, opened 2019-10-09
Waltz torture test with many partitions

whynick1:
1) Added "--num-active-partitions" option to `ClientCli.Validate` cmd; default to 1
2) If more than one active partitions, transactions will be evenly distributed
3) Check serializability each transaction callback with a map of <client_id, <partition_id, high_watermark>>
4) Wait until receive all transaction callback received (from different partitions) before mark success
5) Tested locally

### DIFF
--- a/waltz-test/src/main/python/waltz_ducktape/services/cli/client_cli.py
+++ b/waltz-test/src/main/python/waltz_ducktape/services/cli/client_cli.py
@@ -13,29 +13,30 @@ class ClientCli(Cli):
         """
         super(ClientCli, self).__init__(cli_config_path)
 
-    def validate_txn_cmd(self, partition_id, txn_per_client, num_clients, interval, high_watermark=None):
+    def validate_txn_cmd(self, num_active_partitions, txn_per_client, num_clients, interval, high_watermark=None):
         """
         Return validation cli command to submit and validate transactions, which
         includes validating high water mark, transaction data and optimistic lock.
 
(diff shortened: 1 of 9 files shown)


waltz-test/src/main/python/waltz_ducktape/services/cli/client_cli.py
```diff
@@ -13,29 +13,30 @@ def __init__(self, cli_config_path):
...
-            --txn-per-client <number of transactions per client>
-            --num-clients <number of total clients>
-            --interval <average interval(millisecond) between transactions>
-            --cli-config-path <client cli config file path>
+            --num_active_partitions <number of active partitions> \
+            --txn-per-client <number of transactions per client> \
+            --num-clients <number of total clients> \
+            --interval <average interval(millisecond) between transactions> \
+            --cli-config-path <client cli config file path> \
+            --high-watermark <current high watermark for the partition, default to -1> \
```
criccomini: How does this work when num_active_partitions > 1?

whynick1: Good question, basically it won't work. 
Well, since `--high-watermark` was not used any where yet, we can fix it with a separately, see https://github.com/wepay/waltz/issues/47

criccomini: K

Issue #9, opened 2019-07-08
Make Waltz server StoreSessionImpl.BATCH_SIZE configurable


whynick1:
Based on some ad hoc performance tests, Yasuhiro Matsuda suggested that we might be able to increase producer throughput by increasing the `StoreSessionImpl.BATCH_SIZE`. This is currently hard coded, and we should make it configurable.

abhipardhu:
Merged, https://github.com/wepay/waltz/commit/a8e304d7017f3996218b4ec33071c779381e4b73

PR #54, opened 2019-10-23
Make StoreSessionImpl.BATCH_SIZE configurable in waltz server config

abhipardhu:
Issues: https://github.com/wepay/waltz/issues/9

-- Made StoreSessionImpl.BATCH_SIZE configurable, with default as 100, via waltz server config.
-- Updated corresponding unit tests.

### DIFF
--- a/waltz-server/src/main/java/com/wepay/waltz/server/WaltzServerConfig.java
+++ b/waltz-server/src/main/java/com/wepay/waltz/server/WaltzServerConfig.java
@@ -41,6 +41,11 @@ public class WaltzServerConfig extends AbstractConfig {
     /** Default optimistic lock table size. */
     public static final int DEFAULT_OPTIMISTIC_LOCK_TABLE_SIZE = 30000;
 
+    /** StoreSession batch size, <code>server.storeSessionBatchSize</code> */
+    public static final String STORE_SESSION_BATCH_SIZE = "server.storeSessionBatchSize";
+    /** Default value for {@link #STORE_SESSION_BATCH_SIZE} config. */
+    public static final int DEFAULT_STORE_SESSION_BATCH_SIZE = 100;
+
     /** Feed cache size. */
     public static final String FEED_CACHE_SIZE = "server.feedCacheSize";
     /** Default feed cache size. */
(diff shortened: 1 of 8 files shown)


waltz-server/src/main/java/com/wepay/waltz/server/WaltzServerConfig.java
```diff
@@ -41,6 +41,11 @@
     /** Default optimistic lock table size. */
     public static final int DEFAULT_OPTIMISTIC_LOCK_TABLE_SIZE = 30000;
 
+    /** StoreSession batch size, <code>server.storeSessionBatchSize</code> */
+    public static final String STORE_SESSION_BATCH_SIZE = "server.storeSessionBatchSize";
```
abhipardhu: Better name suggestions are welcome :)

http://thecodelesscode.com/case/220

ymatsuda: `storage.maxBatchSize`?

abhipardhu: Is it okay to do `storage.` in server configs? May be a bit inconsistent with other config keys...

waltz-server/src/test/java/com/wepay/waltz/server/WaltzServerConfigTest.java
```diff
@@ -32,6 +32,7 @@ public void testParsers() {
         map.put(WaltzServerConfig.SERVER_PORT, "8888");
 
         map.put(WaltzServerConfig.OPTIMISTIC_LOCK_TABLE_SIZE, "2000");
+        map.put(WaltzServerConfig.STORE_SESSION_BATCH_SIZE, "100");
```
biplap-sarkar: Can we assert using a size that is different from the default?

waltz-server/src/main/java/com/wepay/waltz/server/WaltzServerConfig.java
ymatsuda: There are `storage.` parameters in the server config already.

abhipardhu: @ymatsuda Totally my bad, I should've looked closely.

Will update the diff.

abhipardhu: Updated.

PR #55, opened 2019-10-23
Do not use leading : in CLASSPATH

wrp:
CLASSPATH should not start with :


wrp:
Just noticed that the leading : gets stripped later, so although this is aesthetically appealing it is not really necessary.

PR #57, opened 2019-10-29
Read config path from System.properties

wrp:
Read path to config file from system.properties, defaulting to /etc/waltz/waltz.cfg.  Can still be overridden with --cli-config-path

### DIFF
--- a/waltz-tools/src/main/java/com/wepay/waltz/tools/client/ClientCli.java
+++ b/waltz-tools/src/main/java/com/wepay/waltz/tools/client/ClientCli.java
@@ -123,7 +123,7 @@ public final class ClientCli extends SubcommandCli {
                     .build();
             Option cfgPathOption = Option.builder("c")
                     .longOpt("cli-config-path")
-                    .desc("Specify client cli config file path")
+                    .desc("Specify client cli config file path (overrides system property waltz.config)")
                     .hasArg()
                     .build();
             Option numActivePartitionOption = Option.builder("ap")
@@ -136,7 +136,7 @@ public final class ClientCli extends SubcommandCli {
             txnPerClientOption.setRequired(true);
             numClientsOption.setRequired(true);
             intervalOption.setRequired(true);
-            cfgPathOption.setRequired(true);
+            cfgPathOption.setRequired(false);
             numActivePartitionOption.setRequired(false);
 
             options.addOption(txnPerClientOption);
@@ -162,7 +162,10 @@ public final class ClientCli extends SubcommandCli {
                 if (avgInterval < 0) {
                     throw new IllegalArgumentException("Found negative: interval must be greater or equals to 0");
                 }
-                String configFilePath = cmd.getOptionValue("cli-config-path");
+                String configFilePath = cmd.getOptionValue(
+                        "cli-config-path",
+                        System.getProperty("waltz.config", "/etc/waltz/waltz.cfg")
+                );
                 WaltzClientConfig waltzClientConfig = getWaltzClientConfig(configFilePath);
 
                 // check optional argument
--- a/docs/waltz-setup.md
+++ b/docs/waltz-setup.md
@@ -27,11 +27,11 @@ Step 1: Use `create` command to create a cluster
 ```
 waltz_uber=/path/to/waltz-uber-X.Y.Z.jar
 java ${waltz_uber+-cp ${waltz_uber}} \
+    -Dwaltz.config=${cli_config_path?} \
     com.wepay.waltz.tools.zk.ZooKeeperCli \
     create \
     --name ${cluster_name?} \
-    --partitions ${partition_count?} \
-    --cli-config-path ${cli_config_path?}
+    --partitions ${partition_count?}
 ```
 
 Step 2: Use `add-storage-node` command to add storage node(s)
--- a/waltz-test/src/main/python/waltz_ducktape/services/cli/client_cli.py
+++ b/waltz-test/src/main/python/waltz_ducktape/services/cli/client_cli.py
@@ -27,12 +27,14 @@ class ClientCli(Cli):
             --num-active-partitions <number of partitions to interact with>
         """
         cmd_arr = [
-            "java -Dlog4j.configuration=file:/etc/waltz-client/waltz-log4j.cfg", self.java_cli_class_name(),
+            "java -Dwaltz.config={} -Dlog4j.configuration=file:/etc/waltz-client/waltz-log4j.cfg".format(
+                self.cli_config_path),
+            self.java_cli_class_name(),
             "validate",
             "--txn-per-client", txn_per_client,
             "--num-clients", num_clients,
             "--interval", interval,
-            "--cli-config-path", self.cli_config_path,
+            "--high-watermark {}".format(high_watermark) if high_watermark is not None else "",
             "--num-active-partitions {}".format(num_active_partitions) if num_active_partitions is not None else ""
         ]
         return self.build_cmd(cmd_arr)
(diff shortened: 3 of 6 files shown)


ymatsuda:
I don't understand the benefit of moving the config file path to a system property? Would you explain it?
And I am afraid having a default config file path make it easy to make mistakes. I'd rather not have a default.

wrp:
The primary benefit of making the config path a system property is that we can access it via standard mechanisms like JAVA_TOOL_OPTIONS.  Rather than having to specify --cli-config-path on every invocation, we can use a -D option in JAVA_TOOL_OPTIONS to get things working.  This is extremely useful in a production environment.  

I have no objection to not using a default, but I don't know a clean way to override the Cli methods to do that reasonably.

ymatsuda:
The config file specifies which waltz cluster it talks to. I think it is dangerous to have a default or to use an environment variable.
JAVA_TOOL_OPTIONS is meant for a JVM options. It seems odd to me to use "-D" there to change the command option.

wrp:
-D is a JVM option. Being able to pass this information in JAVA_TOOL_OPTIONS is the whole point of this change.   If you want to run waltz in a production environment, we need to be able to pass information to the process using standard mechanisms like this.  We control the environment in production.

ymatsuda:
Are you saying --cli-config-path is not usable in production? How come?

wrp:
Having to specify --cli-config-path everywhere is fragile.  It necessitates unnecessary duplication, and violates the principle of DRY.  By being able to specify the config in a single location (eg, a file in /etc/profile.d), is greatly simplifies operations.

ymatsuda:
But if you are managing multiple waltz clusters, having a default is bad. A careless user can damage a cluster easily by using a wrong config file. Having said that, it may be acceptable if the default location is a single place in a system directly like /etc/waltz/ assuming it is protected by permission. A paranoid user like me just won't store any config file there. But I still don't see a necessity to change the command option to the system property.

wrp:
Having mulled this over for a bit, I'm going to disagree strongly.  Consider mysql, rsyslog, nginx, and yum.  When I want to know how these are configured, I look at /etc/my.cnf, /etc/rsyslog.conf, /etc/nginx/conf.d, or /etc/yum.conf.  Although each of those locations is configurable, they are rarely changed.  When they are modified, it causes unnecessary confusion.  Not having a default location will lead to more confusion, not less.  I will now strongly take the position that we *must* have a default location, and discourage using a different path for the configuration.

ymatsuda:
I am not sure what you are disagreeing. I am agreeing to have a single default location. And I am saying that a user must explicitly specify the path using --cli-config-path if they want to place their config in a directory other than the default location.

waltz-tools/src/main/java/com/wepay/waltz/tools/client/ClientCli.java
```diff
@@ -167,7 +167,10 @@ protected void processCmd(CommandLine cmd) throws SubCommandFailedException {
                 if (avgInterval < 0) {
                     throw new IllegalArgumentException("Found negative: interval must be greater or equals to 0");
                 }
-                String configFilePath = cmd.getOptionValue("cli-config-path");
+                String configFilePath = cmd.getOptionValue(
+                        "cli-config-path",
+                        System.getProperty("waltz.config", "/etc/waltz/waltz.cfg")
```
ymatsuda: maybe `/etc/waltz/waltz-client.cfg`. I don't want this to be confused with a server config or a storage config.
Same for other CLI tools.

wrp: I'm not sure I understand how the configs are different.  In our deployments, I am currently using the same config file for all of these things.

whynick1:
```
Having to specify --cli-config-path everywhere is fragile. It necessitates unnecessary duplication, and violates the principle of DRY. By being able to specify the config in a single location (eg, a file in /etc/profile.d), is greatly simplifies operations.
```
I agree we should avoid duplicate config file. When I do integration test there is only one cluster. And I run different CLI command all from the same machine with the *same* config (`--cli-config-path /etc/waltz/waltz-server/cli.cfg`).
```
zookeeper:
    connectString: {{ zk_connect }}
    sessionTimeout: 30000
cluster.root: {{ cluster_root }}
ssl.keyStore.location: {{ ssl_keystore_loc }}
ssl.keyStore.password: {{ ssl_keystore_pwd}}
ssl.trustStore.location: {{ ssl_truststore_loc }}
ssl.trustStore.password: {{ ssl_truststore_pwd }}
```
Based my experience, removing `--cli-config-path` does simply the operation, and seems fine.
```
But if you are managing multiple waltz clusters, having a default is bad. A careless user can damage a cluster easily by using a wrong config file. 
```
That's also a fair concern. One example I can think of, suppose we want to
>1) add a storage0 to cluster0
>2) then, add a storage1 to cluster1

With `--cli-config-path`, I can do it from the same machine, using
>1) `ZookeeperCli add --cli-config-path /etc/waltz/cluster0-cli.cfg`
then,
>2) `ZookeeperCli add --cli-config-path /etc/waltz/cluster1-cli.cfg` 

Note that,  `cluster0-cli.cfg` has `cluster.root: waltz-cluster-0`, and `cluster1-cli.cfg` has `cluster.root: waltz-cluster-1`.

Now Without `--cli-config-path`, we have to either
>1) run two commands from two different machine (with `cli.cfg` contains different `cluster.root`)
or 
>2) run one command, change `/etc/waltz/cli.cfg`, then run another.

If we run the command from machine with "unexpected" cli config, the cluster may get mess up.

@criccomini Any thought?

wrp:
I want to clarify: the primary purpose of this is not to put the path of the config in an environment varialbe.  The primary purpose is to make the config path available from System properties, and an ancillary benefit is that it is then possible to manipulate the value through an environment variable.  In addition, we're adding a default path rather than making the option required.  But you can still override the value using --cli-config-path.

ymatsuda:
If there is neither default or an env variable to set a config path, I would use a bash function like this.
```
zkcli() {
  java -cp /my/class/path com.wepay.waltz.tools.zk.ZooKeeperCli $@ \
    --cli-config-path /my/config/path
}
```
So, I don't have to repeat `--cli-config-path` every time I execute a tool.
If I were managing multiple waltz clusters, I would define a function for each cluster with names that clearly distinguishes the clusters, such as zkcli-prod, storagecli-prod, zkcli-test, storagecli-test.

wrp:
We already use a wrapper script like that, and maintaining it is already becoming a problem.  There is no downside to using a default location.  As I mentioned before, there is great benefit to having consistency.  mysql configs are in /etc/my.cnf.  nginx configs are in /etc/nginx.  yum configs are in /etc/yum.conf.  waltz configs should be in /etc/waltz.

ymatsuda:
I can hardly see a problem of maintaining a script that only sets `--cli-config-path`. It is not doing anything fancy.

wrp:
It is an extra layer of complexity that does not need to exist.  Having a default would simplify running the service in production.  Simplicity is good.  Our job is to make things as simple as possible.  Requiring --cli-config-path is non-standard and imposes an additonal cognitive burden on operators, with no real benefit.

ymatsuda:
You don't type `java -cp /my/class/path com.wepay.waltz.tools.zk.ZooKeeperCli` for every run, do you? I guess you use an alias to save typing. How harder it it to define a function instead of an alias. Not at all, right? Once you set up the function you can forget about `--cli-config-path`.

wrp:
I don't understand your objection to simplifying operations.  If we want waltz to be adopted widely, we should conform to common practices.  Also, we should make it as simple to operate as possible.  Having a default location for configuration is common practice.  Using system properties is common practice.  Forcing every user to write wrapper scripts in order to run waltz in production is counter productive.  It is an additional layer of complexity that introduces the potential for error and has (as far as I can tell) no positive benefit.

PR #59, opened 2019-10-30
Abort bank demo on error

wrp:
I was gettin failures that were hard to track down because docker commands early in the script would fail, but the error messages are obscured because the script blindly continues.  By aborting early on a failure, it is much easier to notice errors.

### DIFF
--- a/bin/docker/container.sh
+++ b/bin/docker/container.sh
@@ -1,4 +1,4 @@
-die() { test -n "$*" && echo "$*"; exit 1; } >&2
+die() { printf "%s${1:+\n}" "$*"; exit 1; } >&2
 
 findImage() {
     # pull or build an image, as necessary


ymatsuda:
Please keep the format of messages of the action result consistent.
I used
.... blah blah blah [some id or name] yada yada yada
Now some are in above format and others.
My preference is to prefix the result messages with "..." and keep ids/names in brackets.

wrp:
I (mostly) reverted the messages.  Moved some to only be before the call, so that it is more clear what has failed if a failure occurs.  Is this what you were thinking?

wrp:
This test needs to fail early (for exampe, if dockerd is not running!).  If the modifications I've made to the comments are a concern, please be specific.

ymatsuda:
Is there any particular reason to switch some of `if [...]` to `if test ...`? The coding style became inconsistent.
I prefer `if [...]` to `if test ...` because the test condition stands out in the syntax.
If you like to use flags like `-z` and `-n`, I suggest using `[ -z ...]` and `[ -n ...]`.

wrp:
No particular reason, just habit.  I've found that the `[` spelling of test has caused a great deal of confusion in its history, and strongly prefer `test`.  But I'll change it back for consistency.  This is not a relevant change here.

git-Bryan (approved):

PR #60, opened 2019-10-30
Cleanup demo-bank-app.sh

wrp:
This is mostly aesthetic.  JAVA_TOOL_OPTIONS is recongnized by the jvm, and there's no need to invoke -cp when CLASSPATH is in the environment.

### DIFF
--- a/bin/demo-bank-app.sh
+++ b/bin/demo-bank-app.sh
@@ -4,14 +4,7 @@ DIR=$(dirname $0)
 
 $DIR/../gradlew --console plain -q copyLibs
 
-CLASSPATH=""
-for file in waltz-demo/build/libs/*.jar;
-do
-    CLASSPATH="$CLASSPATH":"$file"
-done
+export CLASSPATH=$(echo waltz-demo/build/libs/*.jar | tr ' ' :)
+export JAVA_TOOL_OPTIONS=-Dlog4j.configuration=file:config/log4j.properties
 
-JVMOPTS=-Dlog4j.configuration=file:config/log4j.properties
-
-MAINCLASS=com.wepay.waltz.demo.DemoBankApp
-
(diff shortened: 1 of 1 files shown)


ymatsuda:
In general I don't want to export env variables unless it is necessary. I prefer command arguments.

wrp:
I'm not sure I understand the concern.  JAVA_TOOL_OPTIONS and CLASSPATH are standard.    With this change, we ignore the caller's environment.  If anything, this is much cleaner, as not doing this will allow a setting of JAVA_TOOL_OPTIONS in the user's environemnt to modify how the script runs.  If we make this change and override JAVA_TOOL_OPTIONS, the user's environment is less likely to change how the script behaves.

ymatsuda:
Because they are standard environment variables, I'd rather not set it unless it is necessary for better isolation. What if a script calls tools or commands that read these variable? It is always better to be cautious.

wrp:
I still don't understand your concern.  In the current script, if JAVA_TOOL_OPTIONS is set by the caller, the script will behave poorly. (that is, we currently do not have good isolation).  By submitting this change, we clear the environment and get better isolation.

wrp:
IOW, this change gives us better control and is being more cautious, not less.

ymatsuda:
If a user set JAVA_TOOL_OPTIONS to change JVM options and the tool work poorly, that is fine. It is not my concern. My concern is not that a caller may set JAVA_TOOL_OPTIONS but our script is setting it and using JAVA_TOOL_OPTIONS to change our command input. I don't think it is a good practice.

wrp:
But that is *exactly* what the current script is doing with CLASSPATH!

ymatsuda:
Ah- You are right. Sorry, I was totally confused with other PR.

PR #61, opened 2019-11-04
server should unregister partition metrics when unassigned

ymatsuda:
Currently WaltzSever does not unregister partition metrics until shutdown. As the result, a server keep partition metrics after the partition is reassigned to another server.
Metrics were unregistered in `Partition.close()`, but WaltzServer uses `Partition.closeAsync()` when a partition is removed. The fix is to perform unregister in `Partition.closeAsync()`


smithakoduri:
we could have also called `Partition.close()` right instead of `Partition.closeAsync` because `closeAsync` was anyways getting called internally ?

PR #62, opened 2019-11-21
Removed cluster specific configs, key, and numPartitions, from storage configs

abhipardhu:
-- Removed cluster.key and cluster.numPartitions from storage configs.
-- Instead, zk server connection info and cluster.root are added so that the storage can fetch cluster.key and cluster.numPartitions from zk.
-- Moved StoreMetadata related files to waltz-commons.
-- Updated corresponding tests.
-- In local, I made sure that a waltz cluster could be spun up successfully.


ymatsuda:
`com.wepay.waltz.common.metadata.store.internal` is a deep hierarchy. Can we make it `com.wepay.waltz.common.metadata`?
Similarly, `com.wepay.waltz.common.metadata.store.exception` to `com.wepay.waltz.exception`

PR #66, opened 2019-12-03
Implementation of Check Connectivity CLI.

smithakoduri:
Cases Handled:
 - Case 1: All servers and storage nodes are up and running fine.
	Output would be:
	<server_endpoint1>,
		<storage_connect_string1> : true
		<storage_connect_string2> : true
 - Case 2: Storage nodes are down.
	Output would be:
	<server_endpoint1>,
		<storage_connect_string1> : false
		<storage_connect_string2> : false
 - Case 3: Server nodes are unreachable (I.e. if their network connection is down).
	Output would be:
	<server_endpoint1>,
		null

Output format: Map<Endpoint, Map<String, Boolean>>

### DIFF
--- a/waltz-client/src/main/java/com/wepay/waltz/client/internal/InternalBaseClient.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/internal/InternalBaseClient.java
@@ -22,7 +22,9 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
@@ -456,6 +458,16 @@ public abstract class InternalBaseClient implements WaltzNetworkClientCallbacks,
         } catch (Throwable ex) {
             logger.error("failed to set endpoints", ex);
         }
+    }
+
+    /**
+     * Checks the connectivity to the server from the given server endpoint network client.
+     * @param endpoint The server {@link Endpoint}
+     * @return Completable future with the connection status of the given server endpoint.
+     */
+    public CompletableFuture<Optional<Map<String, Boolean>>> checkServerConnectivity(Endpoint endpoint) {
+        WaltzNetworkClient networkClient = getNetworkClient(endpoint);
+        return networkClient.checkServerToStorageConnectivity();
     }
 
 }
--- a/waltz-client/src/main/java/com/wepay/waltz/client/internal/InternalBaseClient.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/internal/InternalBaseClient.java
@@ -22,9 +22,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
@@ -394,7 +392,7 @@ public abstract class InternalBaseClient implements WaltzNetworkClientCallbacks,
         }
     }
 
-    private WaltzNetworkClient getNetworkClient(Endpoint endpoint) {
+    protected WaltzNetworkClient getNetworkClient(Endpoint endpoint) {
         WaltzNetworkClient networkClient = networkClients.get(endpoint);
         if (networkClient == null) {
             networkClient = new WaltzNetworkClient(clientId, endpoint, sslCtx, 0, this, messageProcessingThreadPool);
@@ -458,16 +456,6 @@ public abstract class InternalBaseClient implements WaltzNetworkClientCallbacks,
         } catch (Throwable ex) {
             logger.error("failed to set endpoints", ex);
         }
-    }
-
-    /**
-     * Checks the connectivity to the server from the given server endpoint network client.
-     * @param endpoint The server {@link Endpoint}
-     * @return Completable future with the connection status of the given server endpoint.
-     */
-    public CompletableFuture<Optional<Map<String, Boolean>>> checkServerConnectivity(Endpoint endpoint) {
-        WaltzNetworkClient networkClient = getNetworkClient(endpoint);
-        return networkClient.checkServerToStorageConnectivity();
     }
 
 }
--- a/waltz-client/src/main/java/com/wepay/waltz/client/internal/network/WaltzNetworkClient.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/internal/network/WaltzNetworkClient.java
@@ -5,6 +5,7 @@ import com.wepay.riff.network.MessageProcessingThreadPool;
 import com.wepay.riff.network.NetworkClient;
 import com.wepay.riff.util.Logging;
 import com.wepay.waltz.client.internal.Partition;
+import com.wepay.waltz.common.message.CheckStorageConnectivityRequest;
 import com.wepay.waltz.common.message.FeedRequest;
 import com.wepay.waltz.common.message.HighWaterMarkRequest;
 import com.wepay.waltz.common.message.LockFailure;
@@ -20,6 +21,12 @@ import org.slf4j.Logger;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * A {@link NetworkClient} implementation for waltz clients to communicate with Waltz cluster.
@@ -38,9 +45,12 @@ public class WaltzNetworkClient extends NetworkClient {
     private final MessageProcessingThreadPool messageProcessingThreadPool;
     private final Object lock = new Object();
     private final HashMap<Integer, Partition> partitions;
+    private final AtomicInteger seqNumGenerator = new AtomicInteger(0);
 
     private boolean channelReady = false;
     private volatile boolean running = true;
+    private AtomicReference<CompletableFuture<Optional<Map<String, Boolean>>>> checkConnectivityRef;
+    private AtomicBoolean isConnectivityTriggered;
 
     /**
      * Class Constructor.
@@ -68,10 +78,16 @@ public class WaltzNetworkClient extends NetworkClient {
         this.networkClientCallbacks = networkClientCallbacks;
         this.messageProcessingThreadPool = messageProcessingThreadPool;
         this.partitions = new HashMap<>();
+        this.checkConnectivityRef = new AtomicReference<>();
+        this.isConnectivityTriggered = new AtomicBoolean(false);
     }
 
     /**
-     * Shuts down this instance by un-mounting all partitions and setting {@link #running} to {@code false}.
+     * Shuts down this instance by
+     *  - un-mounting all partitions,
+     *  - setting {@link #running} to {@code false} and
+     *  - completing the {@link #checkConnectivityRef}.future (with empty value) if its not already complete. This
+     *  might be called if the server is down or if there is any network error.
      */
     @Override
     protected void shutdown() {
@@ -83,6 +99,12 @@ public class WaltzNetworkClient extends NetworkClient {
 
                 for (Partition partition : partitions.values()) {
                     partition.unmounted(this);
+                }
+
+                CompletableFuture<Optional<Map<String, Boolean>>> future = checkConnectivityRef.get();
+                if (future != null) {
+                    future.complete(Optional.empty());
+                    this.isConnectivityTriggered.set(false);
                 }
             }
         }
@@ -181,6 +203,33 @@ public class WaltzNetworkClient extends NetworkClient {
         }
     }
 
+    /**
+     * Checks server to storage connectivity.
+     *
+     * Note: There is a possibility that the channel may not be up when this method gets triggered. In that case,
+     * checkServerToStorageConnectivity() gets triggered once the channel is active.
+     *
+     * @return Completable future with the connectivity status to the storage nodes within the cluster.
+     */
+    public CompletableFuture<Optional<Map<String, Boolean>>> checkServerToStorageConnectivity() {
+        ReqId dummyReqId = new ReqId(clientId, 0, 0, seqNumGenerator.incrementAndGet());
+
+        CompletableFuture<Optional<Map<String, Boolean>>> future = checkConnectivityRef.get();
+        if (future != null && isConnectivityTriggered.get()) {
+            return future;
+        }
+
+        if (future == null) {
+            future = new CompletableFuture<>();
+        }
+        if (checkConnectivityRef.compareAndSet(null, future) || !isConnectivityTriggered.get()) {
+            if (sendMessage(new CheckStorageConnectivityRequest(dummyReqId))) {
+                isConnectivityTriggered.set(true);
+            }
+        }
+        return checkConnectivityRef.get();
+    }
+
     @Override
     protected MessageHandler getMessageHandler() {
         WaltzClientHandlerCallbacks handlerCallbacks = new WaltzClientHandlerCallbacksImpl();
@@ -203,6 +252,8 @@ public class WaltzNetworkClient extends NetworkClient {
                 for (Partition partition : partitions.values()) {
                     networkClientCallbacks.onMountingPartition(WaltzNetworkClient.this, partition);
                 }
+
+                checkServerToStorageConnectivity();
             }
         }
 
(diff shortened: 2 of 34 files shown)


wrp:
I haven't looked at the code yet, but have one request.  Could you please change this so that there is no output if the connectivity check passes unless a `-v` or `--verbose` flag is passed.  Succeed quietly, fail loudly.  Scripts are a pain to write if you have to constantly deal with verbose output,

wrp:
ie, return 0 with no output if connectivity is good.  Return 1 and write (to stderr) <storage_connect_string> for each storage node that cannot be reached.  (no output (unless requested with -v) for nodes that are reachable.)

waltz-client/src/main/java/com/wepay/waltz/client/internal/network/WaltzNetworkClient.java
```diff
@@ -181,6 +203,33 @@ public void requestHighWaterMark(ReqId reqId) {
...
+    /**
+     * Checks server to storage connectivity.
+     *
+     * Note: There is a possibility that the channel may not be up when this method gets triggered. In that case,
+     * checkServerToStorageConnectivity() gets triggered once the channel is active.
+     *
+     * @return Completable future with the connectivity status to the storage nodes within the cluster.
+     */
+    public CompletableFuture<Optional<Map<String, Boolean>>> checkServerToStorageConnectivity() {
+        ReqId dummyReqId = new ReqId(clientId, 0, 0, seqNumGenerator.incrementAndGet());
```
ymatsuda: Move this down before `sendMessage()`
Can't we use a dummy seqNum?

waltz-client/src/main/java/com/wepay/waltz/client/internal/network/WaltzNetworkClient.java
```diff
@@ -181,6 +203,33 @@ public void requestHighWaterMark(ReqId reqId) {
...
+        CompletableFuture<Optional<Map<String, Boolean>>> future = checkConnectivityRef.get();
+        if (future != null && isConnectivityTriggered.get()) {
+            return future;
+        }
+
+        if (future == null) {
+            future = new CompletableFuture<>();
+        }
+        if (checkConnectivityRef.compareAndSet(null, future) || !isConnectivityTriggered.get()) {
+            if (sendMessage(new CheckStorageConnectivityRequest(dummyReqId))) {
```
ymatsuda: No need to check the return status of `sendMessage`.
Just do `return future;' after `sendMessage`. Faults will be handled by `InternalRpcClient`.

waltz-client/src/main/java/com/wepay/waltz/client/internal/network/WaltzNetworkClient.java
```diff
@@ -181,6 +203,33 @@ public void requestHighWaterMark(ReqId reqId) {
...
+                isConnectivityTriggered.set(true);
+            }
+        }
+        return checkConnectivityRef.get();
```
ymatsuda: You can't return here because there is a race condition. You need loop until `checkConnectivityRef.get()` returns a future or successfully done `checkConnectivityRef.compareAndSet(null, future)`. See `Partition.getHighWaterMark()`. It is a good example of this pattern.

waltz-client/src/main/java/com/wepay/waltz/client/internal/network/WaltzNetworkClient.java
```diff
@@ -203,6 +252,8 @@ public void onChannelActive() {
                 for (Partition partition : partitions.values()) {
                     networkClientCallbacks.onMountingPartition(WaltzNetworkClient.this, partition);
                 }
+
+                checkServerToStorageConnectivity();
```
ymatsuda: This doesn't work. What we need to do here is to resubmit `CheckStorageConnectivityRequest` if `checkConnectivityRef` has a future.
`checkServerToStorageConnectivity()` **won't** resubmit the request if there **is** a future.

waltz-tools/src/main/java/com/wepay/waltz/tools/client/ClientCli.java
```diff
@@ -521,6 +534,86 @@ protected String getUsage() {
...
+                throw new SubCommandFailedException(String.format("Failed to check topology of the cluster. %n%s",
+                    e.getMessage()));
+            } finally {
+                if (zkClient != null) {
+                    zkClient.close();
+                }
+            }
+        }
+
+        public Set<Endpoint> getListOfServerEndpoints(ZNode root, ZooKeeperClient zkClient) throws Exception {
```
ymatsuda: `public` -> `private`

waltz-tools/src/main/java/com/wepay/waltz/tools/client/ClientCli.java
```diff
@@ -521,6 +534,86 @@ protected String getUsage() {
...
+            for (ServerDescriptor serverDescriptor: serverDescriptors) {
+                serverEndpoints.add(serverDescriptor.endpoint);
+            }
+            return serverEndpoints;
+        }
+
+        private Map<Endpoint, Map<String, Boolean>> checkServerConnections(WaltzClientConfig config,
+                                                                           Set<Endpoint> serverEndpoints) throws Exception {
+            DummyTxnCallbacks callbacks = new DummyTxnCallbacks();
+            WaltzClient client = new WaltzClient(callbacks, config);
```
ymatsuda: Creating a `WaltzClient` for every invocation is heavy. Will you consider using `InternalRpcClient` here?
It is very strange for `WaltzClient` to have `checkServerConnections(Set<Endpoint> serverEndpoints)`.

smithakoduri:
updated code as per review comments.

waltz-client/src/main/java/com/wepay/waltz/client/internal/InternalBaseClient.java
```diff
@@ -458,4 +460,14 @@ private void doSetEndpoints() {
         }
     }
 
+    /**
+     * Checks the connectivity to the server from the given server endpoint network client.
+     * @param endpoint The server {@link Endpoint}
+     * @return Completable future with the connection status of the given server endpoint.
+     */
+    public CompletableFuture<Optional<Map<String, Boolean>>> checkServerConnectivity(Endpoint endpoint) {
```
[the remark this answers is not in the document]
smithakoduri: Created this separate method (that handles single Endpoint) earlier to access the `getNetworkClient` method. 
Changed `getNetworkClient` method to protected (in order to access it in InternalRpcClient) as discussed.

waltz-client/src/main/java/com/wepay/waltz/client/internal/network/WaltzNetworkClient.java
```diff
@@ -38,9 +45,12 @@
     private final MessageProcessingThreadPool messageProcessingThreadPool;
     private final Object lock = new Object();
     private final HashMap<Integer, Partition> partitions;
+    private final AtomicInteger seqNumGenerator = new AtomicInteger(0);
```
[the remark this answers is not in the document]
smithakoduri: ya. I think we can just use 0 instead. fixed.

waltz-client/src/main/java/com/wepay/waltz/client/internal/network/WaltzNetworkClient.java
```diff
@@ -38,9 +45,12 @@
...
 
     private boolean channelReady = false;
     private volatile boolean running = true;
+    private AtomicReference<CompletableFuture<Optional<Map<String, Boolean>>>> checkConnectivityRef;
+    private AtomicBoolean isConnectivityTriggered;
```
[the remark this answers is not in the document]
smithakoduri: fixed.

waltz-client/src/main/java/com/wepay/waltz/client/internal/network/WaltzNetworkClient.java
```diff
@@ -84,6 +100,12 @@ protected void shutdown() {
                 for (Partition partition : partitions.values()) {
                     partition.unmounted(this);
                 }
+
+                CompletableFuture<Optional<Map<String, Boolean>>> future = checkConnectivityRef.get();
+                if (future != null) {
+                    future.complete(Optional.empty());
+                    this.isConnectivityTriggered.set(false);
```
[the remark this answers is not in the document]
smithakoduri: fixed.

waltz-server/src/main/java/com/wepay/waltz/server/WaltzServer.java
```diff
@@ -352,6 +353,10 @@ public boolean isClosed() {
         return partitionIds;
     }
 
+    public NetworkServer getNetworkServer() {
```
ymatsuda: I don't want to expose `NetworkServer`. Can you change it to `closeNetworkServer()`?

waltz-tools/src/main/java/com/wepay/waltz/tools/client/ClientCli.java
```diff
@@ -521,6 +536,88 @@ protected String getUsage() {
         }
     }
 
+    private static final class CheckConnectivity extends Cli {
```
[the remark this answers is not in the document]
smithakoduri: hmmm ya. Makes sense. will create ClusterCli.

waltz-server/src/main/java/com/wepay/waltz/server/WaltzServer.java
smithakoduri: will fix.

waltz-common/src/main/java/com/wepay/waltz/common/message/MessageCodecV0.java
```diff
@@ -7,6 +7,9 @@
 import com.wepay.waltz.common.util.Utils;
 import com.wepay.waltz.exception.RpcException;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public class MessageCodecV0 implements MessageCodec {
```
ymatsuda: Not necessary to update V0 codec. This is for communicating with old servers.

smithakoduri:
> I haven't looked at the code yet, but have one request. Could you please change this so that there is no output if the connectivity check passes unless a `-v` or `--verbose` flag is passed. Succeed quietly, fail loudly. Scripts are a pain to write if you have to constantly deal with verbose output,

The current idea behind this cli implementation is to use this for displaying it on the dashboard. We are currently working on Waltz metrics and figuring out how/what to display on the dashboard that would be useful for debugging later.

waltz-common/src/main/java/com/wepay/waltz/common/message/MessageCodecV0.java
smithakoduri: aahh ok. fixed.

wrp:
If you wish to use the command like this for displaying information on a dashboard, that is the correct time to invoke the verbose option.  If you want the tool to be generally useful for operational purposes, please make the default behavior be silent.  Excessive verbosity is a bug, not a feature.  "Succeed quietly, fail loudly" should always be a design feature of a command line tool.

biplap-sarkar:
Although I agree with "succeed quietly, fail loudly" sentiment in general, I am not sure if that covers all the cases here where network topology changes silently. If a server goes down cleanly, zk entry for that server will be removed as well. The command returning 0 doesn't necessarily mean that the topology ping worked. It might also mean that the topology ping worked, but with a slightly changed topology.

wrp:
In general, operationally, I want a tool that answers the question "is the
cluster usable".  I'd like something that simply returns 0 if it is
(whether degraded or not) or non-zero if it is not.  If we want additional
detail on degraded states, I would prefer that be available by request.
(eg, by specifying --verbose)

Please don't make the operator specify `--quiet` or redirect stdout to
/dev/null just to make the simple query.  Scripts become cluttered quickly
when hacks are necessary to suppress unnecessary output.

ymatsuda:
The cluster health is not binary. For example, the cluster is still usable when not all storage nodes are reachable as long as we can establish the quorum. I am not sure of the usefulness of this cli if it succeeds quietly when the cluster is merely "usable". It is not a good thing to leave degraded state uncorrected for long. So, I think the success should mean every thing is good not just the cluster is usable.
I'd like the default behavior to be informational.

criccomini:
We need to be clear on what use cases we have, and what use cases we're solving for with this particular tool and PR.

@wrp, you have clearly outlined a use case that you are solving for: a human running a CLI command that determines whether the cluster can receive reads and writes for all partitions, regardless of overall health. I believe you want to run this command when you first start up a cluster, right? If so, why wouldn't you want to know if all server and storage nodes are fully functional, rather than just alive enough to have quorum?

There are other use cases, too:

1. Allow operators to see the current status of all server and storage nodes when diagnosing a problem.
2. Provide a tool that monitoring tools can use to check for liveness
3. Provide a tool that metrics tools can use to generate metrics

Can we agree on which use cases this PR is solving for before discussing behavior?

wrp:
I think that having verbose ouptput is a perfectly reasonable use case, but i would like the verbose output to be dependent on a flag.  Eg, `--verbose`.  I  prefer that the default be non-verbose, but would not object terribly if the output is suppressed via `--quiet`.  however, I do not want the user to have to suppress output via redirections.

biplap-sarkar:
> We need to be clear on what use cases we have, and what use cases we're solving for with this particular tool and PR.

The use case serving as an objective for this PR is to solve for use cases 2 and 3 with 1 being a side effect. 
Additionally, cluster health is not 0 or 1. It is possible that a cluster is still usable but has lost a few nodes. The goal for this PR is to find out different problematic scenarios that might have different SLAs for remediation rather than to find out if the cluster is up or not. 

If the name of the cli is causing confusion, we don't have to call it `check-connectivity` and rather call it `cluster-status`

criccomini:
@wrp, can you please respond to my questions, though? You responded by talking about behavior again. I'm talking about use cases.

criccomini:
> If the name of the cli is causing confusion, we don't have to call it check-connectivity and rather call it cluster-status

I also agree that the naming might be part of the confusion. Still, I feel like @wrp is just fundamentally solving for a different thing: a tool that he runs specifically when he spins up a new cluster that simply tells him whether the cluster was spun up successfully. In that case, an exit(0) seems fine, but that's not what you're talking about. I'm trying to clarify what he's solving for.

wrp:
At the moment, the only use case I'm thinking of is answering the question "Is the cluster usable?"  It may very well be theoretically impossible to answer that question, since you cannot determine whether a future operation will succeed but at the present time I am unaware of a reasonable mechanism for answering the question.  I just want a healthcheck.  If we are writing a tool that queries the cluster, I am  expressing a preference that it not be verbose, and that its verbosity be controlled by a run-time flag defaulting to brevity.  If this is a tool that is querying the cluster and periodically writing metrics to a tsdb, then it should not be providing any output other than the metrics anyway.  Perhaps the naming is an issue, as I am certainly thinking of this as a status check.  But really I'm just asking that --verbose not be the default.

wrp:
Waltz already has metrics exposed on jmx and on the jetty port that we can scrape and track in a tsdb, so we don't need a tool in waltz to do the scraping (IOW, use case 3 is already solved).  In general, a tool that produces textual output that needs to be read and interpreted by a human is not particularly useful; humans will misinterpret the data even if they have the specialized knowledge required to correctly interpret it.  Being more verbose doesn't solve that problem, but makes it worse.  If you want to track "connectivity", I suppose what you're wanting to do is add a metric on the server that records heartbeats to a storage node.   So to solve use case 2, you just need to record "liveness" (I put this in quotes because I don't know exactly the definiton of that term in this context) and expose that on the jetty port.   But I didn't intend to stir up a can of worms when I first commented on this PR.  I just want `--quiet` to be the default!

biplap-sarkar:
"Is the cluster usable" does not have a boolean answer. If we force to make it boolean, I am not sure what's the value of that answer for that question. If the concern is that humans will misinterpret the data, a boolean answer is already a misinterpretation. Cluster running in degraded mode is a practical scenario rather than being a theoretical one and based on the scenario, the priority of the escalation will be different. I am not sure if there is any critical scenario where a boolean answer would be sufficient for any operational purpose. 
Waltz does have jmx metrics, but those are for individual servers and storage nodes. I am not sure how current metrics can be used to create a coherent picture of the cluster.
Regarding the tool directly emitting metrics instead of generating any stdout content is a valid comment. But we have an internal document regarding the metrics story (which also outlines that approach) and it might make sense to have that discussion over there instead of this PR. As mentioned before, use case 1 is a by-product of this PR but it seems strange to intentionally get rid of that.

criccomini:
Talked a bit offline with @wrp.

First, I agree in general with the unix principal to not print output when a command is successful. However, some commands are MEANT to do that (e.g. `ls`). That's the whole point.

Second, I think there's disagreement about what "working" and "healthy" mean. Do they mean you can read/write to all partitions? Or does it mean that all servers and storage nodes are fully up with all partitions properly assigned?

Here's where I come down. There are three things we want:

1. Verify (strict). Verifies that all servers and storage nodes are up, can communicate with one another, and that all partitions are assigned to servers and storage nodes.
2. Verify (quorum). Verifies that all partitions are assigned and can receive reads and writes. 
3. Print current status to console for human inspection.

IMO, (1) and (2) should be a single command. (3) should be a separate command. I believe that verify command should be silent by default, and exit(0) when successful. I believe the print command should not be quiet--printing stuff is the reason it exists.

We can argue about whether verify should default to strict or quorum.  IMO, it should be paranoid, and default to strict with a `--quorum` switch that allows users to relax the verification. I believe this is less dangerous to the user. See also Kafka's list of CLI commands, many of which expose information that's also exposed via metrics. The CLI commands are still useful.

@wrp asked if (3) should exist at all, since we expose this information through JMX and metrics--users can put these metrics into a dashboard. I believe we should still have a CLI command for out-of-box experience, and because in reality people don't hook up all their metrics from day one. Having a CLI tool eases the transition to production-readiness. See also 

From a "color" perspective, I would say:

1. verify strict = true, green
2. verify strict = false, verify quorum = true, yellow
3. verify strict = false, verify quorum = false, red

PR #67, opened 2019-12-04
use cached client high-water mark for performance

ymatsuda:
This PR improves the performance of `WaltzClient.submit()` using cached client high-water marks instead of calling `getClientHighWaterMark()` callback which reads the high-water mark from an application database.

### DIFF
--- a/waltz-client/src/main/java/com/wepay/waltz/client/AbstractClientCallbacksForJDBC.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/AbstractClientCallbacksForJDBC.java
@@ -97,8 +97,6 @@ public abstract class AbstractClientCallbacksForJDBC implements WaltzClientCallb
             try {
                 Connection connection = dataSource.getConnection();
                 try {
-                    connection.setAutoCommit(true);
-
                     // Read the high-water mark from database and update the cache .
                     return selectHighWaterMark(partitionId, connection);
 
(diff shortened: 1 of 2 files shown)


waltz-client/src/main/java/com/wepay/waltz/client/AbstractClientCallbacksForJDBC.java
```diff
@@ -97,8 +97,6 @@ public long getClientHighWaterMark(int partitionId) {
             try {
                 Connection connection = dataSource.getConnection();
                 try {
-                    connection.setAutoCommit(true);
```
ymatsuda: No need to alter auto-commit setting here.

waltz-client/src/main/java/com/wepay/waltz/client/internal/InternalStreamClient.java
```diff
@@ -90,7 +90,7 @@ public TransactionBuilderImpl getTransactionBuilder(TransactionContext context)
         Partition partition = getPartition(context.partitionId(numPartitions));
         partition.ensureMounted();
 
-        return new TransactionBuilderImpl(partition.nextReqId(), callbacks.getClientHighWaterMark(partition.partitionId));
+        return new TransactionBuilderImpl(partition.nextReqId(), partition.clientHighWaterMark());
```
ymatsuda: `partition.clientHighWaterMark()` reads the current client high-water mark from `AtomicLong` in the partition object. It is incremented immediately after `applyTransaction()`.

PR #77, opened 2020-01-16
a new lock type

ymatsuda:
This PR introduces the third type of lock, the APPEND lock. (I'd like to solicit a better name!)
It doesn't check the lock high-water mark for the lock compatibility, but it always updates the lock high-water mark. Here is the summary of three lock types.

WRITE    => checks the lock high-water mark and updates the lock high-water mark
READ      => checks the lock high-water mark but doesn't update the lock high-water mark
APPEND => doesn't check the lock high-water mark but always updates the lock high-water mark

APPEND lock alone may not seem so useful. It is meant to be used with other locks.

Suppose you have a hierarchical resource (parent/child). You want to update the child resource in rapid succession, and you don't want other processes to update the child while doing it. Using a write lock on the parent for every child update is inefficient. It causes self-imposed lock contentions. One solution is to mark the parent resource with the ownership. It tells who is owning the resource. This transaction uses a write lock. A child update transaction uses a read lock to ensure that the current process still owns the resource. It will fail when the ownership is taken by another process.

This seems to work. But there is a potential issue. There is no way for the new owner to ensure that the child update is in-flight when taking the ownership. The new owner may commit a transaction changing the ownership before a child update become visible to it.

An APPEND lock can be used when this is a concern. The ownership change transaction takes WRITE locks on the parent resource and the child resource. The child update transaction takes a READ lock on the parent and an APPEND lock on the child. This ensures that child updates are not causing lock contentions, and there is no in-flight transactions when the ownership changes.

### DIFF
--- a/waltz-server/src/main/java/com/wepay/waltz/server/internal/Locks.java
+++ b/waltz-server/src/main/java/com/wepay/waltz/server/internal/Locks.java
@@ -19,7 +19,7 @@ public class Locks {
     private final BitSet locks;
     private final int numHashFuncs;
 
-    private int numActiveLocks = 0;
+    private int numActiveLockRequests = 0;
 
     /**
      * Class contsructor.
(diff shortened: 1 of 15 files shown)


waltz-server/src/main/java/com/wepay/waltz/server/internal/Locks.java
```diff
@@ -37,15 +37,16 @@ public Locks(int size, int numHashFuncs, long defaultHighWaterMark) {
...
      */
     public boolean begin(LockRequest request) {
         synchronized (locks) {
-            if (begin(request.writeLocks) && begin(request.readLocks)) {
-                // Mark entries for write locks
+            if (begin(request.writeLocks) && begin(request.readLocks) && begin(request.appendLocks)) {
+                // Mark entries for write locks and append locks
                 mark(request.writeLocks);
-                numActiveLocks++;
+                mark(request.appendLocks);
```
criccomini: Nit: somewhat subtle that readLocks doesn't need to be marked but appendLocks does. perhaps worth a comment.

PR #78, opened 2020-01-22
Making metric group configurable (for server only for now). 

biplap-sarkar:
This PR is just an example of the relevant changes if we want to use this approach. Even though the local tests work after this, the problem introduced by this is we'll have to make metric group configurable in cli's as well.


biplap-sarkar:
Discussed the option of getting metrics from the server instead of jetty port. We'll still make metric group configurable but cli client won't be needing any explicit option. Instead, the server will have a map of all metric group with metric data and before returning metric response to the cli, the server can filter out the response for the configured metric group only.

PR #86, opened 2020-02-24
Include server-storage connectivity checks in cluster-cli verify

abhipardhu:
-- Included server-storage connectivity checks in cluster-CLI verify.
-- In the verify command, only the connectivity statuses that correspond to the replicas of the partitions that a server owns are considered.
-- Updated WaltzNetworkClient to use the same pattern for sending all CLI requests.
-- Suppressed casting warnings in WaltzNetworkClient.
-- Updated unit tests for ClusterCli.

### DIFF
--- a/waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
+++ b/waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
@@ -389,10 +389,10 @@ public final class ClusterCli extends SubcommandCli {
                                 .validationResultsMap
                                 .put(
                                     ValidationResult.ValidationType.SERVER_STORAGE_CONNECTIVITY,
-                                    getStorageConnectivityValidationResult(
+                                    buildStorageConnectivityValidationResult(
                                         partitionIdToReplicas.get(partition.partitionId),
-                                        endpoint,
-                                        storageConnectivityResults
+                                        storageConnectivityResults,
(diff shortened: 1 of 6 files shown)


waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
```diff
@@ -187,37 +192,78 @@ protected void processCmd(CommandLine cmd) throws SubCommandFailedException {
...
+                verifyValidation(
+                    ValidationResult.ValidationType.PARTITION_ASSIGNMENT_ZK_SERVER_CONSISTENCY,
+                    partitionsValidationResultList
+                );
+
+                verifyValidation(
+                    ValidationResult.ValidationType.SERVER_STORAGE_CONNECTIVITY,
+                    partitionsValidationResultList
+                );
+            } catch (RuntimeException e) {
```
smithakoduri: isn't this exception also covered in `catch (Exception e)` in the else case? do we need this (assuming we don't want to print any specific error corresponding to it.) ??

abhipardhu: yes it is. But, I was getting a findbugs warning that a generic Exception is caught :(

smithakoduri: ok ok, got it. let it be then.

PR #87, opened 2020-03-02
Implement storage control file parser

abhipardhu:
-- Implemented a `waltz-storage.ctl` control file parser to dump control file information to the console.
-- `DiskCli dump-control-file <control-file-path>`, with `-partition` option to dump a specific PartitionInfo. `-assign` option to dump all assigned partitions' PartitionInfos.
-- This is helpful if we have to read a control file without needing to attach the disk to a running waltz-storage node. (If attached, `StorageCli list ...` should give the control file information)
-- Renamed SegmentCli to DiskCli to extend it to parse control files.
-- Other minor updates as needed.
-- Added unit tests.


waltz-tools/src/main/java/com/wepay/waltz/tools/storage/disk/DiskCli.java
```diff
@@ -294,17 +295,134 @@ private void reportStatistic() {
...
+                    throw new SubCommandFailedException("'assigned' is supported only if dumping for all partitions");
+                }
+
+                initWithControlFilePath(controlFilePath);
+
+                List<List<String>> content = new ArrayList<>();
+                if (cmd.hasOption("p")) {
+                    int partitionId = Integer.parseInt(cmd.getOptionValue("p"));
+                    int numPartitions = controlFile.getNumPartitions();
+                    if (partitionId >= numPartitions) {
```
patrickghannage: Do we want to check for negative inputs as well?

PR #88, opened 2020-03-04
Enforce host:port in Waltz ZooKeeperCli#Add

zzlbuaa:
- Add "host:port" format enforcement for add-storage-node (Add) command in ZooKeeperCli
- Add unit test for that


waltz-tools/src/test/java/com/wepay/waltz/tools/zk/ZooKeeperCliTest.java
```diff
@@ -259,6 +259,50 @@ public void testAddExistingStorageNodeShouldFail() throws Exception {
...
+            ZNode clusterRootZNode = new ZNode(CLUSTER_ROOT);
+            ZNode storeRoot = new ZNode(clusterRootZNode, StoreMetadata.STORE_ZNODE_NAME);
+            ZNode groupNode = new ZNode(storeRoot, StoreMetadata.GROUP_ZNODE_NAME);
+            ZooKeeperClient zkClient = new ZooKeeperClientImpl(connectString, ZK_SESSION_TIMEOUT);
+            ZooKeeperCli.Create.createCluster(zkClient, clusterRootZNode, CLUSTER_NAME, NUM_PARTITIONS);
+            ZooKeeperCli.Create.createStores(zkClient, clusterRootZNode, NUM_PARTITIONS, STORAGE_GROUP_MAP, STORAGE_CONNECTION_MAP);
+
+            int prevGroupSize = zkClient.getData(groupNode, GroupDescriptorSerializer.INSTANCE).value.groups.size();
+
+            // add mis-formatted storage node should fail
```
smithakoduri: may be you can just add the below code to the `testAddStorageNode()` itself.

smithakoduri: Actually, get a second opinion on that. Looks like each test case in ZooKeeperCli is handling only one specific scenario. So, may be having a separate test case should be fine.

zzlbuaa: > may be you can just add the below code to the `testAddStorageNode()` itself.
Yes that makes sense, but I was thinking that there's like testAddExistingStorageNodeShouldFail under ZookerperCliTest and those tests seem fine-grained so I still added a new unit test. But I'm unsure which is the better way to go..

waltz-tools/src/main/java/com/wepay/waltz/tools/zk/ZooKeeperCli.java
```diff
@@ -535,6 +535,11 @@ protected void processCmd(CommandLine cmd) throws SubCommandFailedException {
                 int adminPort = Integer.parseInt(cmd.getOptionValue("storage-admin-port"));
                 int groupId = Integer.parseInt(cmd.getOptionValue("group"));
 
+                String[] storageHostAndPort = storage.split(":");
+                if (storageHostAndPort.length != 2) {
+                    throw new IllegalArgumentException("Storage must be in format of host:admin_port");
```
abhipardhu: I think it should be in host:port, we have -a for admin port.

zzlbuaa: > I think it should be in host:port, we have -a for admin port.

Ohh yes, I'll modify that

zzlbuaa:
Modified `host:admin_port` to `host:port` in the error message. PTAL @abhipardhu @smithakoduri @patrickghannage

PR #89, opened 2020-03-06
Add getNumPartitions() for WaltzClient

zzlbuaa:
Add getNumPartition() to get total number of partitions of a WaltzClient, so that application clients can use this number to calculate which partition id a transaction goes to when the transaction is submitted, and it's easier for debugging partition-related issues.


zzlbuaa:
Can I get a review on this plz? Thanks😄 @abhipardhu @patrickghannage @smithakoduri @ymatsuda

PR #93, opened 2020-03-27
Create new StoreSessions on replica assignment changes

abhipardhu:
1. On replica assignment change watcher events, close old store sessions and create new sessions on replica assignment changes.
2. Creating new sessions will trigger recovery which results in updates to replicastates info in the ZooKeeper.
3. Earlier, this was done on the next write. However, if writes are not happening, then multiple replica assignment changes without recovery could make the cluster unrecoverable in the future.
4. Since replica assignment changes are done manually, it is hard to enforce "Wait until recovery after every assignment change".
5. So, this pull request attempts to make it automatic by eagerly opening new sessions. It always compares the new assignment with the existing assignment, and further processing is done only if there are any changes. This is to handle no-op watcher handler invocations that happen when there are connection issues between the client and the ZK server.

Tested on a local cluster.

### DIFF
--- a/waltz-server/src/main/java/com/wepay/waltz/store/internal/StoreImpl.java
+++ b/waltz-server/src/main/java/com/wepay/waltz/store/internal/StoreImpl.java
@@ -1,22 +1,32 @@
 package com.wepay.waltz.store.internal;
 
 import com.wepay.riff.util.Logging;
+import com.wepay.waltz.common.util.DaemonThreadFactory;
 import com.wepay.waltz.server.WaltzServerConfig;
 import com.wepay.waltz.store.Store;
 import com.wepay.waltz.store.StorePartition;
+import com.wepay.waltz.store.exception.GenerationMismatchException;
+import com.wepay.waltz.store.exception.RecoveryFailedException;
 import com.wepay.waltz.store.exception.StoreException;
 import com.wepay.waltz.common.metadata.ReplicaAssignments;
 import com.wepay.waltz.common.metadata.ReplicaId;
 import com.wepay.waltz.common.metadata.StoreMetadata;
 import com.wepay.waltz.common.metadata.StoreParams;
+import com.wepay.waltz.store.exception.StoreSessionManagerException;
+import com.wepay.zktools.zookeeper.NodeData;
 import com.wepay.zktools.zookeeper.WatcherHandle;
 import com.wepay.zktools.zookeeper.ZNode;
 import com.wepay.zktools.zookeeper.ZooKeeperClient;
 import org.slf4j.Logger;
 
+import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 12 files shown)


waltz-server/src/test/java/com/wepay/waltz/store/internal/StoreImplTest.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Map;
+import java.util.Properties;
+import java.util.stream.Collectors;
+
+public class StoreImplTest {
+    private static final int NUM_REPLICAS = 3;
+    private static final int CLUSTER_NUM_PARTITIONS = 3;
+
+    @Test
+    public void testReplicaAssignmentChangeEffects() throws Exception {
```
abhipardhu: Looking for suggestions/thoughts on the testing front...

This is more of an integration test than a unit test, please let me know if it should be at a different place...

waltz-storage/src/main/resources/log4j.xml
```diff
@@ -7,7 +7,7 @@
         <param name="Target" value="System.out" />
         <param name="Encoding" value="UTF-8" />
         <layout class="org.apache.log4j.PatternLayout">
-            <param name="ConversionPattern" value="%d{yyyy-MM-dd HH:mm:ss.SSS} UTC %-5p: %c - %m%n" />
+            <param name="ConversionPattern" value="[%t]%d{yyyy-MM-dd HH:mm:ss.SSS} UTC %-5p: %c - %m%n" />
```
abhipardhu: Unrelated to what this PR is trying to solve...
Just a generic change to help in better logging on local clusters...

waltz-test/src/main/java/com/wepay/waltz/test/util/IntegrationTestHelper.java
```diff
@@ -46,6 +46,9 @@
 public class IntegrationTestHelper {
 
     private static final int STORAGE_GROUP_ID = 0;
+    // Use smaller caches in testing.
```
biplap-sarkar: Sorry for nitting again. But can you provide some reasoning in comments (such as using the defaults ends up tests needing much more memory). Just wanted to make sure that the time and effort you spent in order to figure out the problem is preventing someone new not doing the same thing.

waltz-server/src/main/java/com/wepay/waltz/store/internal/StoreImpl.java
```diff
@@ -111,4 +124,47 @@ public ConnectionConfig getConnectionConfig() {
...
+        }
+    }
+
+    private StoreSession getStoreSession(StoreSessionManager storeSessionManager) {
+        StoreSession storeSession = null;
+
+        try {
+            storeSession = storeSessionManager.getStoreSession();
+        } catch (GenerationMismatchException | StoreSessionManagerException | RecoveryFailedException ex) {
+            logger.warn("Failed to get StoreSession", ex);
```
patrickghannage, line 166: nit: Do we want to log more info on which storeSession failed to be created ?

patrickghannage: Probably taken care by "ex"

PR #94, opened 2020-03-30
Updated StorageCli.SyncPartitionAssignments 

smithakoduri:
Un-assign partitions that are no longer part of storage node as per ZooKeeper metadata.

### DIFF
--- a/waltz-tools/src/main/java/com/wepay/waltz/tools/storage/StorageCli.java
+++ b/waltz-tools/src/main/java/com/wepay/waltz/tools/storage/StorageCli.java
@@ -36,6 +36,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


waltz-tools/src/main/java/com/wepay/waltz/tools/storage/StorageCli.java
```diff
@@ -780,10 +782,33 @@ private void syncPartitionAssignments(String cliConfigPath) throws Exception {
...
+                        for (int partitionId : partitionIds) {
+                            if ((assignedPartitionStatus.get(partitionId) == null) || !assignedPartitionStatus.get(partitionId)) {
+                                futures.add(storageAdminClient.setPartitionAssignment(partitionId, true, false));
+                            }
+                            assignedPartitionStatus.remove(partitionId);
+                        }
+
+                        // Remove the remaining old partitions that are no longer to be handled.
+                        for (int partitionId : assignedPartitionStatus.keySet()) {
+                            storageAdminClient.setPartitionAssignment(partitionId, false, false).get();
```
zzlbuaa, line 805: Could we do a *futures.add(storageAdminClient.setPartitionAssignment(partitionId, false, false));* here and wait only once in the *CompletableFuture.allOf.get()*?

PR #95, opened 2020-04-03
Add ZooKeeper connect timeout for CLIs

zzlbuaa:
There's an issue with CLIs that: when ZooKeeper is not running, attempts to run several Storage/ZooKeeper/Cluster CLI commands will get stuck and never exit because the initiation of ZooKeeperClientImpl will keep retrying in this process.

This PR adds a ZOOKEEPER_CONNECT_TIMEOUT parameter to all such initiation calls. Then, if ZooKeeper is not running, such CLI command will exit and throw errors.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ project.ext {
     slf4jVersion = '1.7.25'
     mockitoVersion = '1.9.5'
     assertjVersion = '3.8.0'
-    zkToolsVersion = '0.7.0'
+    zkToolsVersion = '0.7.1'
     yamlVersion = '1.20'
     riffVersion = '2.4.3'
     jacksonVersion = '2.9.6'
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ project.ext {
     slf4jVersion = '1.7.25'
     mockitoVersion = '1.9.5'
     assertjVersion = '3.8.0'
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -20,7 +20,7 @@ project.ext {
     slf4jVersion = '1.7.25'
     mockitoVersion = '1.9.5'
     assertjVersion = '3.8.0'
-    zkToolsVersion = '0.6.0'
+    zkToolsVersion = '0.7.0'
```
zzlbuaa, line 23: ZooKeeperClientImpl() is updated in 0.7.0:
https://github.com/wepay/zktools/commit/62cd5ccabdceaa5abebb1c57e8c65e2073d2fc1f#comments

build.gradle
```diff
@@ -20,7 +20,7 @@ project.ext {
     slf4jVersion = '1.7.25'
     mockitoVersion = '1.9.5'
     assertjVersion = '3.8.0'
-    zkToolsVersion = '0.6.0'
+    zkToolsVersion = '0.7.1'
```
zzlbuaa, line 23: Updated with new version of zktools: https://github.com/wepay/zktools/commit/c498f88cba887318320ec2ccb8f7c6b31c94d787

zzlbuaa: @ymatsuda Can I get a review for this new change? Thanks!

PR #96, opened 2020-04-09
Check recovery finished in ZKCli

zzlbuaa:
This PR is to check for each replica session if the recovery is completed and add it to zkCli.list output, so that we can better know the current state of the cluster.

### DIFF
--- a/waltz-tools/src/main/java/com/wepay/waltz/tools/zk/ZooKeeperCli.java
+++ b/waltz-tools/src/main/java/com/wepay/waltz/tools/zk/ZooKeeperCli.java
@@ -123,13 +123,15 @@ public final class ZooKeeperCli extends SubcommandCli {
                     for (int id = 0; id < storeParams.numPartitions; id++) {
                         ZNode znode = new ZNode(partitionRoot, Integer.toString(id));
                         Map<ReplicaId, ReplicaState> replicaState;
+                        Long sessionId = null;
                         try {
                             PartitionMetadata partitionMetadata = zkClient.getData(znode, PartitionMetadataSerializer.INSTANCE).value;
                             replicaState = partitionMetadata.replicaStates;
+                            sessionId = partitionMetadata.sessionId;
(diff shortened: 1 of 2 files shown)


waltz-tools/src/main/java/com/wepay/waltz/tools/zk/ZooKeeperCli.java
```diff
@@ -186,13 +188,19 @@ private void listClusterInfoAndServerPartitionAssignments(ZooKeeperClient zkClie
...
+        private void listReplicaState(ZNode znode, Map<ReplicaId, ReplicaState> replicaState, Long storeSessionId) {
             System.out.println("store [" + znode + "] replica states:");
 
             if (replicaState.size() > 0) {
                 Map<ReplicaId, ReplicaState> sortedReplicaState = new TreeMap<>(replicaState);
                 for (Map.Entry<ReplicaId, ReplicaState> entry : sortedReplicaState.entrySet()) {
-                    System.out.println("  " + entry.getKey() + ", SessionId: " + entry.getValue().sessionId + ", closingHighWaterMark: " + ((entry.getValue().closingHighWaterMark) == ReplicaState.UNRESOLVED ? "UNRESOLVED" : entry.getValue().closingHighWaterMark));
+                    long replicaSessionId = entry.getValue().sessionId;
+                    long closingHighWaterMark = entry.getValue().closingHighWaterMark;
+                    boolean recoveryFinished = storeSessionId != null && replicaSessionId == storeSessionId && closingHighWaterMark == ReplicaState.UNRESOLVED;
```
zzlbuaa, line 199: In the recovery process, the sessionId is first updated in partitionMetadata at the beginning, and each replica session's sessionId is updated to the latest sessionId after the recovery finished, and its closing HWM set to UNRESOLVED. So here I check for these two conditions for complete.

zzlbuaa:
We want to check if recovery is finished in clusterCli.verify instead of zkCli. ZKCli is only for list metadata in zookeeper. Will close this PR.

PR #99, opened 2020-04-15
Code to support/handle partition option of verify CLI.

smithakoduri:
Fix include:
 - Handling Verify Cli for a single partition if "partition" option is present
 - Setting "PARTITION_ASSIGNMENT_ZK_SERVER_CONSISTENCY" to failed state if,
    - a partition is handled by multiple servers, or
    - if a partition is not handled by any server.
 - Wrap "validationResultsMap" inside a synchronized map.

### DIFF
--- a/waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
+++ b/waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
@@ -259,7 +259,11 @@ public final class ClusterCli extends SubcommandCli {
                     partitionToStorageStatusMap, partitionsValidationResultList);
 
                 // Update validation result for remaining partitions if not already updated.
-                updateRemainingValidationResults(consistencyValidationFuture, connectivityValidationFuture,
+                updateRemainingValidationResults(consistencyValidationFuture,
+                    ValidationResult.ValidationType.PARTITION_ASSIGNMENT_ZK_SERVER_CONSISTENCY,
+                    partitionsValidationResultList);
+                updateRemainingValidationResults(connectivityValidationFuture,
+                    ValidationResult.ValidationType.SERVER_STORAGE_CONNECTIVITY,
                     partitionsValidationResultList);
 
                 // Verify all results or just one partition if "partition" option is present
@@ -282,32 +286,20 @@ public final class ClusterCli extends SubcommandCli {
             }
         }
 
-        private void updateRemainingValidationResults(CompletableFuture<Void> consistencyValidationFuture,
-                                                      CompletableFuture<Void> connectivityValidationFuture,
+        private void updateRemainingValidationResults(CompletableFuture<Void> future,
+                                                      ValidationResult.ValidationType validationType,
                                                       List<PartitionValidationResults> partitionsValidationResultList) throws Exception {
             String error = "No Server Endpoint found.";
 
             try {
-                consistencyValidationFuture.get(TIMEOUT_IN_SECONDS, TimeUnit.SECONDS);
+                future.get(TIMEOUT_IN_SECONDS, TimeUnit.SECONDS);
             } catch (ExecutionException | TimeoutException exception) {
                 error = exception.getMessage();
             }
 
             failMissingValidationResults(
                 partitionsValidationResultList,
-                ValidationResult.ValidationType.PARTITION_ASSIGNMENT_ZK_SERVER_CONSISTENCY,
-                error
-            );
-
-            try {
-                connectivityValidationFuture.get(TIMEOUT_IN_SECONDS, TimeUnit.SECONDS);
-            } catch (ExecutionException | TimeoutException exception) {
-                error = exception.getMessage();
-            }
-
-            failMissingValidationResults(
-                partitionsValidationResultList,
-                ValidationResult.ValidationType.SERVER_STORAGE_CONNECTIVITY,
+                validationType,
                 error
             );
         }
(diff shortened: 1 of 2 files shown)


waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
```diff
@@ -737,6 +771,14 @@ private void buildZookeeperPartitionAssignmentsValidation(ClusterManager cluster
                                 error);
                     PartitionValidationResults partitionValidationResults = partitionValidationResultsList.get(partitionInfo.partitionId);
                     partitionValidationResults.validationResultsMap.put(validationResult.type, validationResult);
+
+                    if (failServerPartitionConsistency) {
+                        ValidationResult serverConsistencyValidationResult = new ValidationResult(
+                            ValidationResult.ValidationType.PARTITION_ASSIGNMENT_ZK_SERVER_CONSISTENCY,
+                            ValidationResult.Status.FAILURE, error);
+                        partitionValidationResults.validationResultsMap.put(serverConsistencyValidationResult.type, serverConsistencyValidationResult);
+
```
patrickghannage, line 780: nit: extra space

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
```diff
@@ -737,6 +771,14 @@ private void buildZookeeperPartitionAssignmentsValidation(ClusterManager cluster
                                 error);
                     PartitionValidationResults partitionValidationResults = partitionValidationResultsList.get(partitionInfo.partitionId);
                     partitionValidationResults.validationResultsMap.put(validationResult.type, validationResult);
+
```
patrickghannage, line 774: I was thinking instead of failing the zk_server test in here, we should probably fail it in buildServerZKPartitionAssignmentsValidation just after `assignmentMatchMap.forEach((partitionId, match) -> {`. For the partitionId, we just check if it's zk_test has failed in step1, if yes we directly fail it's zk_server_test. That way we don't have to use putIfAbsent, we can blindly put (overwrite) the failure message multiple time.

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
```diff
@@ -767,7 +809,8 @@ protected String getUsage() {
 
             PartitionValidationResults(int partitionId) {
                 this.partitionId = partitionId;
-                this.validationResultsMap = new EnumMap<>(ValidationResult.ValidationType.class);
+                this.validationResultsMap =
+                    Collections.synchronizedMap(new EnumMap<>(ValidationResult.ValidationType.class));
```
patrickghannage, line 813: synchronizedMap makes sure threads have an up-to-date view of all buckets plus allows null values. Is that why we chose it? It's probably not a big deal but I am curious why ConcurrentHashMap was not of choice here.

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
```diff
@@ -289,6 +282,36 @@ protected void processCmd(CommandLine cmd) throws SubCommandFailedException {
             }
         }
 
+        private void updateRemainingValidationResults(CompletableFuture<Void> consistencyValidationFuture,
```
patrickghannage, line 285: We don't have to do it but it's just an idea. We could have the method take a Map<Future,ValidationType>> to have it more generic. So can then loop instead of type the same code twice below.

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
smithakoduri: I was thinking to only fail server_partition_consistency if zk_test failed because of "partition handled by multiple server" error. Now I am thinking it shouldn't matter. Will make the change as you suggested.

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
smithakoduri: Because I wanted to keep the map as a EnumMap. As we have a synchronizedMap that we can wrap around any other Map, I thought this can be used. 

If we are not very particular to use EnumMap then we can instead change to ConcurrentHashMap.

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
smithakoduri: makes sense. 👍

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
patrickghannage: Ow I missed that... Yeah... Forget what I said, we won't have that many threads modifying a partition map. We have only 5 keys in our enum :)

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
```diff
@@ -211,6 +214,13 @@ protected void processCmd(CommandLine cmd) throws SubCommandFailedException {
...
                 int numPartitions = clusterManager.numPartitions();
 
+                if (cmd.hasOption("partition")) {
+                    pId = Integer.parseInt(cmd.getOptionValue("partition"));
+                    if ((pId < 0) || (pId >= numPartitions)) {
+                        throw new IllegalArgumentException("Partition " + pId + " is not valid.");
+                    }
+                }
+
                 for (int partitionId = 0; partitionId < clusterManager.numPartitions(); partitionId++) {
```
abhipardhu, line 224: naming nit: It is better to change this for-local variable to pId and use partitionId for the variable with larger scope.

PR #100, opened 2020-04-22
VerifyCli: Add verification for recovery complete

zzlbuaa:
This PR is to verify if the recover is complete for each partition in VerifyCli.
Added a new ValidationType.REPLICA_RECOVERY_STATUS, and this validation result is set to success if and only if:
1. The set of replicas in replica assignments is the same as the set of replicas in replica states.
2. For each replica, the session id in the replicaStates equals the session id in partitionMetadata, and closingHighWaterMark is UNRESOLVED.

If recovery not complete for some replica, the validation result will be printed like:
```
Validation STORAGE_RECOVERY_COMPLETE failed for partition 0
Validation error is:
Storage waltz-storage:55280 recovery not completed

### DIFF
--- a/waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
+++ b/waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
@@ -267,8 +267,11 @@ public final class ClusterCli extends SubcommandCli {
                 // Step5: Validate recovery is completed and replica states updated in ZK
                 ZNode partitionRoot = new ZNode(storeRoot, StoreMetadata.PARTITION_ZNODE_NAME);
 
+                Map<Integer, Set<String>> zkPartitionToStorageNodeSetMap =
+                    getZkPartitionToStorageNodeSetMapping(storeMetadata);
+
                 buildStorageRecoveryCompleteValidationResult(numPartitions, partitionRoot,
-                    storeMetadata.getReplicaAssignments(), zkClient, partitionsValidationResultList);
+                    zkPartitionToStorageNodeSetMap, zkClient, partitionsValidationResultList);
 
                 // Update validation result for remaining partitions if not already updated.
                 updateRemainingValidationResults(consistencyValidationFuture,
@@ -386,25 +389,18 @@ public final class ClusterCli extends SubcommandCli {
          *
          * @param numPartitions Total number of partitions in the cluster.
          * @param partitionRoot ZNode used to fetch replica states for each partition.
-         * @param replicaAssignments Replica assignments for all the partitions.
+         * @param zkPartitionToStorageNodeSetMap Map of partitionId to set of storageConnectString.
          * @param zkClient ZooKeeper Client used to fetch partition metadata.
          * @param partitionsValidationResultsList a {@code List<PartitionValidationResults>} that contains all the
          *                                       validation results.
          */
         private void buildStorageRecoveryCompleteValidationResult(int numPartitions,
                                                                   ZNode partitionRoot,
-                                                                  ReplicaAssignments replicaAssignments,
+                                                                  Map<Integer, Set<String>> zkPartitionToStorageNodeSetMap,
                                                                   ZooKeeperClient zkClient,
                                                                   List<PartitionValidationResults> partitionsValidationResultsList) {
-            // A set that contains all replica ids in replica assignments
-            Set<ReplicaId> assignmentReplicaIdSet = new HashSet<>();
-            for (Map.Entry<String, int[]> entry : replicaAssignments.replicas.entrySet()) {
-                for (int partitionId : entry.getValue()) {
-                    assignmentReplicaIdSet.add(new ReplicaId(partitionId, entry.getKey()));
-                }
-            }
-
             for (int partitionId = 0; partitionId < numPartitions; partitionId++) {
+                Set<String> storageNodeSet = zkPartitionToStorageNodeSetMap.get(partitionId);
                 String error = "";
                 ZNode zNode = new ZNode(partitionRoot, Integer.toString(partitionId));
                 Map<ReplicaId, ReplicaState> replicaState;
@@ -421,13 +417,13 @@ public final class ClusterCli extends SubcommandCli {
                 }
 
                 for (Map.Entry<ReplicaId, ReplicaState> entry : replicaState.entrySet()) {
-                    ReplicaId replicaId = entry.getKey();
-                    if (!assignmentReplicaIdSet.contains(replicaId)) {
+                    String storageNodeConnectString = entry.getKey().storageNodeConnectString;
+                    if (!storageNodeSet.contains(storageNodeConnectString)) {
                         error = error.concat(System.lineSeparator())
-                            .concat(String.format("%s replica info missing in replica assignments", replicaId));
+                            .concat(String.format("Storage %s replica info missing in replica assignments", storageNodeConnectString));
                         continue;
                     }
-                    assignmentReplicaIdSet.remove(replicaId);
+                    storageNodeSet.remove(storageNodeConnectString);
 
                     boolean recoveryComplete =
                         (sessionId != null)
@@ -435,8 +431,12 @@ public final class ClusterCli extends SubcommandCli {
                         && (entry.getValue().closingHighWaterMark == ReplicaState.UNRESOLVED);
                     if (!recoveryComplete) {
                         error = error.concat(System.lineSeparator())
-                            .concat(String.format("%s recovery not completed", replicaId));
+                            .concat(String.format("Storage %s recovery not completed", storageNodeConnectString));
                     }
+                }
+
+                if (storageNodeSet.isEmpty()) {
+                    zkPartitionToStorageNodeSetMap.remove(partitionId);
                 }
 
                 ValidationResult validationResult = new ValidationResult(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java [hunk 1]
```diff
@@ -364,6 +375,52 @@ private void buildStoragePartitionValidationResult(int numPartitions,
...
+                    replicaState = partitionMetadata.replicaStates;
+                    sessionId = partitionMetadata.sessionId;
+                } catch (KeeperException | ZooKeeperClientException e) {
+                    replicaState = Collections.emptyMap();
+                    sessionId = null;
+                    error = "Cannot obtain replica states";
+                }
+
+                for (Map.Entry<ReplicaId, ReplicaState> entry : replicaState.entrySet()) {
+                    boolean recoveryComplete =
```
zzlbuaa, line 399: In the recovery process, the sessionId is first updated in partitionMetadata at the beginning, and each replica session's sessionId and closing HWM won't be updated until the recovery finished. So here I check for these two conditions for the complete.

waltz-tools/src/test/java/com/wepay/waltz/tools/cluster/ClusterCliTest.java
```diff
@@ -184,6 +184,10 @@ public void testVerifyCommand() throws Exception {
             assertTrue(outContent.toString("UTF-8")
                 .contains("Validation PARTITION_QUORUM_STATUS failed for partition 2"));
 
+            // Check that STORAGE_RECOVERY_COMPLETE didn't fail
+            assertFalse(outContent.toString("UTF-8").contains("Recovery not completed"));
+            assertFalse(outContent.toString("UTF-8").contains("Cannot obtain replica states"));
+
```
zzlbuaa, line 190: Didn't find a good way to add unit test for the scenario when recovery is ongoing and this verification fails, so I just added here to check STORAGE_RECOVERY_COMPLETE is successful in normal case. Suggestions are welcomed on how to do the testing.

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
```diff
@@ -816,7 +873,8 @@ public String toString() {
                 PARTITION_ASSIGNMENT_ZK_SERVER_CONSISTENCY,
                 SERVER_STORAGE_CONNECTIVITY,
                 PARTITION_ASSIGNMENT_ZK_STORAGE_CONSISTENCY,
-                PARTITION_QUORUM_STATUS
+                PARTITION_QUORUM_STATUS,
+                STORAGE_RECOVERY_COMPLETE
```
smithakoduri, line 877: nit: naming convention. May be `Replica_recovery_status` or something instead?

smithakoduri:
Otherwise, LGTM.

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
zzlbuaa: updated based on suggestion, PTAL😄

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
abhipardhu: Does it cover the case where recovery didn't start at all?

zzlbuaa: > Does it cover the case where recovery didn't start at all?

@abhipardhu No, I don't think so.. I think the verification in this PR only checks for if the replicas are in `recovery completed` or `recovery ongoing` status.
To check if certain recovery was triggered/completed, the verification needs to know the sessionId before the recovery to compare with, but we didn't take that parameter here. 
Do you think we need to add such option here? Like add a `-s` parameter for sessionId, and verify if all replicas have an updated sessionId compared to that input?

smithakoduri: @abhipardhu : Are you referring to the case where replica assignment of a storage node is updated with a partition_Id, but that replica is not part of the corresponding partition's replica state?

For eg: In the below scenario, the replica states doesn't show `storage04:300`.

store [/waltz/store] replica and group assignments:
  storage01:300 = [0], GroupId: 1
  storage02:300 = [0] GroupId: 2
  storage03:300 = [0], GroupId: 0
  storage04:300 = [0], GroupId: 0
store [/waltz/store] connections:
  storage01:300 has admin port: 301
  storage02:300 has admin port: 301
  storage03:300 has admin port: 301
store [/waltz/store/partition/0] replica states:
  ReplicaId(0,storage01:300), SessionId: 2, closingHighWaterMark: UNRESOLVED
  ReplicaId(0,storage02:300), SessionId: 2, closingHighWaterMark: UNRESOLVED
  ReplicaId(0,storage03:300), SessionId: 2, closingHighWaterMark: UNRESOLVED

zzlbuaa: @abhipardhu @smithakoduri Add verification for consistency between `replicaAssignments` vs `replicaStates`, plz take a 👀

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
```diff
@@ -364,6 +378,103 @@ private void buildStoragePartitionValidationResult(int numPartitions,
...
+                    PartitionMetadata partitionMetadata = zkClient.getData(zNode, PartitionMetadataSerializer.INSTANCE).value;
+                    replicaState = partitionMetadata.replicaStates;
+                    sessionId = partitionMetadata.sessionId;
+                } catch (KeeperException | ZooKeeperClientException e) {
+                    replicaState = Collections.emptyMap();
+                    sessionId = null;
+                    error = "Cannot obtain replica states";
+                }
+
+                Map<ReplicaId, ReplicaState> sortedReplicaState = new TreeMap<>(replicaState);
```
smithakoduri, line 425: any particular reason you are creating a new `sortedReplicaState`?

zzlbuaa: The only purpose here is to keep the error msg ordered by ReplicaId when they are being added.

zzlbuaa: Also suggested by @abhipardhu on this. Update it and remove it since ordering by `storageNodeConnectString` won't improve the readability of error msg, and msgs are already grouped by partitionId.

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
```diff
@@ -364,6 +378,103 @@ private void buildStoragePartitionValidationResult(int numPartitions,
...
+        private void buildStorageRecoveryCompleteValidationResult(int numPartitions,
+                                                                  ZNode partitionRoot,
+                                                                  ReplicaAssignments replicaAssignments,
+                                                                  ZooKeeperClient zkClient,
+                                                                  List<PartitionValidationResults> partitionsValidationResultsList) {
+            // A set that contains all replica ids in replica assignments
+            Set<ReplicaId> assignmentReplicaIdSet = new TreeSet<>();
+            for (Map.Entry<String, int[]> entry : replicaAssignments.replicas.entrySet()) {
+                for (int partitionId : entry.getValue()) {
+                    assignmentReplicaIdSet.add(new ReplicaId(partitionId, entry.getKey()));
```
smithakoduri, line 403: Instead of creating `new ReplicaId's` (not sure if its a right way to do it), I was wondering if it would be better if you can store the replica assignment mapping in a different way. For eg. `Map<Integer,Set<String>>` where key can be the partitionId and value can be replicas of that partition. Or any other different way. 

You can make use of `getZkPartitionToStorageNodeMapping(...)` method as well, and also for simplicity you can change the return type from `Map<Integer, List<String>>` to `Map<Integer, Set<String>>`.

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
smithakoduri: Ya, order shouldn't matter.

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
```diff
@@ -454,6 +564,19 @@ private StorageAdminClient openStorageAdminClient(String storageHost, Integer st
             return zkPartitionToReplicaMap;
         }
 
+        private Map<Integer, Set<String>> getZkPartitionToStorageNodeSetMapping(StoreMetadata storeMetadata) throws Exception {
```
smithakoduri, line 567: There seems to be two methods now for `getZkPartitionToStorageNode` Mapping: 
 - One with return type as `Map<Integer, Set<String>>`
 - Other with return type as `Map<Integer, List<String>>`

Only make use of one method and do the corresponding changes.

smithakoduri:
@zzlbuaa: Can you make one more change. I am seeing a small bug in `buildStoragePartitionValidationResult(...)` method.

Can you replace 
`if (zkPartitionToStorageNodeMap.getOrDefault(id, new ArrayList<>()).size() != partitionToStorageStatusMap.getOrDefault(id,new HashMap<>()).size()) {` with the below lines. I seem to have missed out to check if the parameters match or not.

```
                Set<String> zkStorageNodeMapPerId = zkPartitionToStorageNodeMap.getOrDefault(id,
                    new HashSet<>());
                Map<String, Boolean> storageStatusMapPerId = partitionToStorageStatusMap.getOrDefault(id,
                    new HashMap<>());
                if ((zkStorageNodeMapPerId.size() != storageStatusMapPerId.size()) || (!zkStorageNodeMapPerId.equals(storageStatusMapPerId.keySet()))){
```

waltz-tools/src/main/java/com/wepay/waltz/tools/cluster/ClusterCli.java
zzlbuaa: Removed the method with `Map<Integer, List<String>>`, thanks for suggestion :)

PR #102, opened 2020-06-09
fix append lock in mock client

ymatsuda:
This fixes Append locks support in the mock client. Append locks were ignored before this fix.


wrp:
LGTM

PR #109, opened 2020-07-10
ducktape test - client disconnected from server

hrdlotom:
WaltzClient establishes a network connection with Waltz Server to send/receive transactions. This test simulates network disconnection between client and server.

How to test it:
add print statements connection_interruption.py to see when command to drop/allow traffic through a server port is fired. With this knowledge check the metrics output of server and storage nodes to see that transactions aren't going through. 
Further testing can me done with tcpdump port 6060/7070 or by checking iptables

### DIFF
--- a/waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption.py
+++ b/waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption.py
@@ -7,31 +7,31 @@ class ConnectionInterruption(Thread):
     A scheduler to drop incoming traffic to a port periodically.
     """
 
-    def __init__(self, service, interruption_length, no_of_interruptions, service_node_index):
+    def __init__(self, service, interruption_length, num_of_interruptions, service_node_index):
         """
             Construct a new 'ConnectionInterruption' object.
 
             :param service: The service having incoming traffic to a port blocked
             :param interruption_length: The interval(milliseconds) between two interruptions and the length of each interruption
-            :param no_of_interruptions: Total number of network connection interruptions
+            :param num_of_interruptions: Total number of network connection interruptions
             :param service_node_index: A service node used for this test
         """
         Thread.__init__(self)
         self.service = service
         self.interruption_length = interruption_length
-        self.interruptions = no_of_interruptions
+        self.num_of_interruptions = num_of_interruptions
         self.service_node_index = service_node_index
 
     def run(self):
         node = self.service.nodes[self.service_node_index]
-        for interruption in range(self.interruptions):
+        for interruption in range(self.num_of_interruptions):
             # let waltz server receive transactions as usual
-            sleep(self.interruption_length/1000)
+            sleep(self.interruption_length)
 
             # disable connection on port
             self.service.logger.info("Closing a port")
             node.account.ssh_capture("sudo iptables -I INPUT -p tcp --destination-port {} -j DROP".format(self.service.port))
-            sleep(self.interruption_length / 1000)
+            sleep(self.interruption_length)
 
             # enable connection on port
             self.service.logger.info("Opening a port")
--- a/waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption.py
+++ b/waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption.py
@@ -1,38 +1,70 @@
-from threading import Thread
+from waltz_ducktape.tests.produce_consume_validate import ProduceConsumeValidateTest
+from ducktape.mark.resource import cluster
+from ducktape.mark import parametrize
+from ducktape.cluster.cluster_spec import ClusterSpec
+from ducktape.utils.util import wait_until
 from time import sleep
+from random import randrange
 
 
-class ConnectionInterruption(Thread):
+class ConnectionInterruption(ProduceConsumeValidateTest):
     """
-    A scheduler to drop incoming traffic to a port periodically.
+        Class for tests simulating network issues between server and storage.
     """
 
-    def __init__(self, service, interruption_length, num_of_interruptions, service_node_index):
-        """
-            Construct a new 'ConnectionInterruption' object.
+    MIN_CLUSTER_SPEC = ClusterSpec.from_list([
+        {'cpu': 1, 'mem': '1GB', 'disk': '25GB', 'additional_disks': {'/dev/sdb': '100GB'}, 'num_nodes': 3},
+        {'cpu': 1, 'mem': '3GB', 'disk': '15GB', 'num_nodes': 2},
+        {'cpu': 1, 'mem': '1GB', 'disk': '25GB', 'num_nodes': 1}])
 
-            :param service: The service having incoming traffic to a port blocked
-            :param interruption_length: The interval(milliseconds) between two interruptions and the length of each interruption
-            :param num_of_interruptions: Total number of network connection interruptions
-            :param service_node_index: A service node used for this test
-        """
-        Thread.__init__(self)
-        self.service = service
-        self.interruption_length = interruption_length
-        self.num_of_interruptions = num_of_interruptions
-        self.service_node_index = service_node_index
+    def __init__(self, test_context):
+        super(ConnectionInterruption, self).__init__(test_context=test_context)
 
-    def run(self):
-        node = self.service.nodes[self.service_node_index]
-        for interruption in range(self.num_of_interruptions):
-            # let waltz server receive transactions as usual
-            sleep(self.interruption_length)
+    @cluster(cluster_spec=MIN_CLUSTER_SPEC)
+    @parametrize(num_active_partitions=1, txn_per_client=100, num_clients=2, interval=600, timeout=360,
+                 network_interruption_length=10, num_interruptions=3, duration_between_interruptions=5)
+    @parametrize(num_active_partitions=4, txn_per_client=100, num_clients=2, interval=1000, timeout=300,
+                 network_interruption_length=20, num_interruptions=1, duration_between_interruptions=20)
+    def test_produce_consume_with_network_interruption(self, num_active_partitions, txn_per_client, num_clients, interval, timeout,
+                                                network_interruption_length, num_interruptions, duration_between_interruptions):
+        validation_cmd = self.client_cli.validate_txn_cmd(num_active_partitions, txn_per_client, num_clients, interval)
+        node_idx = self.get_server_node_idx(randrange(min(num_active_partitions, len(self.waltz_server.nodes))))
+        self.run_produce_consume_validate(lambda: self.produce_consume_with_network_interruption(validation_cmd, timeout, network_interruption_length, num_interruptions, duration_between_interruptions, node_idx, num_active_partitions))
+
+    def produce_consume_with_network_interruption(self, validation_cmd, timeout, network_interruption_length, num_interruptions, duration_between_interruptions, node_idx, num_active_partitions):
+        """
+        Set up waltz and interrupt network between a waltz client node and a server node.
+
+        :param validation_cmd: The command that is send to ClientCli
+        :param timeout: Test timeout
+        :param network_interruption_length: Interval in seconds during which client won't be able to connect to server
+        :param num_interruptions: Number of connection interruption cycles
+        :param duration_between_interruptions: Interval in seconds that represents duration between network interruptions
+        :param node_idx: Index of a server node to which client won't be able to connect
+        :param num_active_partitions: Number of active partitions
+        """
+
+        # Start waltz cluster and wait until a storage node registers first transaction
+        self.verifiable_client.start(validation_cmd)
+        admin_port = self.waltz_storage.admin_port
+        port = self.waltz_storage.port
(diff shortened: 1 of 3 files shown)


waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption.py
```diff
@@ -0,0 +1,38 @@
...
+        self.service = service
+        self.interruption_length = interruption_length
+        self.interruptions = no_of_interruptions
+        self.service_node_index = service_node_index
+
+    def run(self):
+        node = self.service.nodes[self.service_node_index]
+        for interruption in range(self.interruptions):
+            # let waltz server receive transactions as usual
+            sleep(self.interruption_length/1000)
```
biplap-sarkar, line 29: It seems like we are accepting `interruption_length` as milliseconds, but we are dropping the relevant precision and are converting it to seconds. Any specific reason why we are setting an expectation that the time is in ms. Can we change the param to accept seconds instead?

waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption.py
```diff
@@ -0,0 +1,38 @@
...
+
+            # disable connection on port
+            self.service.logger.info("Closing a port")
+            node.account.ssh_capture("sudo iptables -I INPUT -p tcp --destination-port {} -j DROP".format(self.service.port))
+            sleep(self.interruption_length / 1000)
+
+            # enable connection on port
+            self.service.logger.info("Opening a port")
+            node.account.ssh_capture("sudo iptables -D INPUT -p tcp --destination-port {} -j DROP".format(self.service.port))
```
biplap-sarkar, line 38: Seems like the review tool is complaining about the absent of a newline at the end of the file. Can you add a newline?

waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption.py
hrdlotom: I believe that python doesn't convert the number to an integer like Java, but a decimal point number instead. But I can definitely change interruption_length to be represented in seconds. Interval between transactions is represented in milliseconds, on the other hand timeout is in seconds and I feel like there won't be a need to set _interruption_length_ to let's say 1.5s and 1s or 2s seconds will do as well.

waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption.py
hrdlotom: 👍

waltz-test/src/main/python/waltz_ducktape/tests/validation/smoke_test.py
```diff
@@ -44,6 +45,15 @@ def test_produce_consume_while_killing_a_server_node(self, num_active_partitions
         self.run_produce_consume_validate(lambda: self.simple_validation_func(validation_cmd, timeout),
                                           lambda: self._kill_a_server_node(num_active_partitions))
 
+    @cluster(cluster_spec=MIN_CLUSTER_SPEC)
+    @parametrize(num_active_partitions=1, txn_per_client=80, num_clients=2, interval=600, timeout=300, interruption_length=5000, no_interruptions=4)
```
biplap-sarkar, line 49: We should be consistent in terms of variable naming. We have `num_active_partitions` and `num_clients` but then we have `no_interruptions` as well, which I am assuming is passing the number of interruptions.

smithakoduri:
It would be better to create a single "network_interruption_test.py" that covers all the cases for both client-server and server-storage connections instead of creating separate classes for both the cases (The other PR for server-storage network interruption test seems to be creating a new Class). That way we can arrange all scenarios of the same functionality test under one file.

waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption.py
```diff
@@ -0,0 +1,38 @@
...
+        self.service = service
+        self.interruption_length = interruption_length
+        self.num_of_interruptions = num_of_interruptions
+        self.service_node_index = service_node_index
+
+    def run(self):
+        node = self.service.nodes[self.service_node_index]
+        for interruption in range(self.num_of_interruptions):
+            # let waltz server receive transactions as usual
+            sleep(self.interruption_length)
```
smithakoduri, line 29: What does "interruption_length" refer to? Is it the amount of time that the network connection will be down?
If yes, then what is the reason behind this statement? 
May be we can add a separate parameter to provide the frequency of the interruption and use that variable here?

waltz-test/src/main/python/waltz_ducktape/tests/validation/smoke_test.py
```diff
@@ -44,6 +45,15 @@ def test_produce_consume_while_killing_a_server_node(self, num_active_partitions
...
                                           lambda: self._kill_a_server_node(num_active_partitions))
 
+    @cluster(cluster_spec=MIN_CLUSTER_SPEC)
+    @parametrize(num_active_partitions=1, txn_per_client=100, num_clients=2, interval=600, timeout=360, interruption_length=10, num_interruptions=3)
+    @parametrize(num_active_partitions=4, txn_per_client=100, num_clients=2, interval=1000, timeout=300, interruption_length=20, num_interruptions=1)
+    def test_produce_consume_client_disconnects(self, num_active_partitions, txn_per_client, num_clients, interval, timeout,
+                                                interruption_length, num_interruptions):
+        validation_cmd = self.client_cli.validate_txn_cmd(num_active_partitions, txn_per_client, num_clients, interval)
+        self.run_produce_consume_validate(lambda: self.simple_validation_func(validation_cmd, timeout),
+                                          lambda: self._disconnect_and_reconnect_client(num_active_partitions, interruption_length, num_interruptions))
```
smithakoduri, line 55: The other PR for server-storage network interruption test seems to be creating a new Class "connection_interruption_storage.py" that includes both the test case and the network interruption logic.

In order to be consistent and to make it easier I would recommend having a separate Class "network_interruption_test.py" or "connection_interruption_test.py” that covers all the scenarios/tests corresponding to the network interruption functionality. So, it will contain test cases for "client-server" and "server-storage" connections under one single file.

waltz-test/src/main/python/waltz_ducktape/tests/validation/smoke_test.py
```diff
@@ -44,6 +45,15 @@ def test_produce_consume_while_killing_a_server_node(self, num_active_partitions
         self.run_produce_consume_validate(lambda: self.simple_validation_func(validation_cmd, timeout),
                                           lambda: self._kill_a_server_node(num_active_partitions))
 
+    @cluster(cluster_spec=MIN_CLUSTER_SPEC)
+    @parametrize(num_active_partitions=1, txn_per_client=100, num_clients=2, interval=600, timeout=360, interruption_length=10, num_interruptions=3)
+    @parametrize(num_active_partitions=4, txn_per_client=100, num_clients=2, interval=1000, timeout=300, interruption_length=20, num_interruptions=1)
+    def test_produce_consume_client_disconnects(self, num_active_partitions, txn_per_client, num_clients, interval, timeout,
```
smithakoduri, line 51: NIT: `test_produce_consume_while_network_interruption` or `test_produce_consume_while_connection_interruption` instead of `test_produce_consume_client_disconnects `

waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption.py
```diff
@@ -0,0 +1,38 @@
...
+
+            # disable connection on port
+            self.service.logger.info("Closing a port")
+            node.account.ssh_capture("sudo iptables -I INPUT -p tcp --destination-port {} -j DROP".format(self.service.port))
+            sleep(self.interruption_length)
+
+            # enable connection on port
+            self.service.logger.info("Opening a port")
+            node.account.ssh_capture("sudo iptables -D INPUT -p tcp --destination-port {} -j DROP".format(self.service.port))
```
smithakoduri, line 38: You can move the both of these iptables logic to a separate helper functions so that you can use them in the server-storage connection test scenario as well (in order to avoid redundancy).

waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @cluster(cluster_spec=MIN_CLUSTER_SPEC)
+    @parametrize(num_active_partitions=1, txn_per_client=100, num_clients=2, interval=600, timeout=360,
+                 interruption_duration=10, num_interruptions=3, delay_between_interruptions=5)
+    @parametrize(num_active_partitions=4, txn_per_client=100, num_clients=2, interval=1000, timeout=300,
+                 interruption_duration=20, num_interruptions=1, delay_between_interruptions=20)
+    def test_produce_consume_with_network_interruption(self, num_active_partitions, txn_per_client, num_clients, interval, timeout,
+                                                interruption_duration, num_interruptions, delay_between_interruptions):
+        validation_cmd = self.client_cli.validate_txn_cmd(num_active_partitions, txn_per_client, num_clients, interval)
+        self.run_produce_consume_validate(lambda: self.produce_consume_with_network_interruption(validation_cmd, timeout, interruption_duration, num_interruptions, delay_between_interruptions, num_active_partitions, interval))
```
smithakoduri, line 31: NIT: line is too long. Let's split it into multiple lines.
Same applies for other lines as well.

waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def drop_traffic_to_port(self, node, port):
+        node.account.ssh_capture("sudo iptables -I INPUT -p tcp --destination-port {} -j DROP".format(port))
+
+    def enable_traffic_to_port(self, node, port):
+        node.account.ssh_capture("sudo iptables -D INPUT -p tcp --destination-port {} -j DROP".format(port))
+
+    def produce_consume_with_network_interruption(self, validation_cmd, timeout, interruption_duration, num_interruptions, delay_between_interruptions, num_active_partitions, interval):
```
smithakoduri, line 39: NIT: rename to mention that this function is referring to client-server network interruption.

waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption_test.py
```diff
@@ -0,0 +1,81 @@
...
+        {'cpu': 1, 'mem': '1GB', 'disk': '25GB', 'additional_disks': {'/dev/sdb': '100GB'}, 'num_nodes': 3},
+        {'cpu': 1, 'mem': '3GB', 'disk': '15GB', 'num_nodes': 2},
+        {'cpu': 1, 'mem': '1GB', 'disk': '25GB', 'num_nodes': 1}])
+
+    def __init__(self, test_context):
+        super(ConnectionInterruptionTest, self).__init__(test_context=test_context)
+
+    @cluster(cluster_spec=MIN_CLUSTER_SPEC)
+    @parametrize(num_active_partitions=1, txn_per_client=100, num_clients=2, interval=600, timeout=360,
+                 interruption_duration=10, num_interruptions=3, delay_between_interruptions=5)
```
smithakoduri, line 25: nit: `interrupt_duration` instead of `interruption_duration` may be?? what do you think?

waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption_test.py
```diff
@@ -0,0 +1,81 @@
...
+        storage = self.get_host(self.waltz_storage.nodes[0].account.ssh_hostname, admin_port)
+        wait_until(lambda: self.is_max_transaction_id_updated(storage, port, partition, -1), timeout_sec=timeout)
+
+        node = self.waltz_server.nodes[node_idx]
+        for interruption in range(num_interruptions):
+            sleep(delay_between_interruptions)
+            # disable connection on port
+            self.drop_traffic_to_port(node, self.waltz_server.port)
+
+            #verify that new transactions aren't processed
```
smithakoduri, line 68: NIT: space after `#`
Add a command to mentioned that we choose to sleep for `interval` amount of time to make sure that all transactions that are sent to server (by the client) until then have persisted in the replicas.

waltz-test/src/main/python/waltz_ducktape/tests/validation/connection_interruption_test.py
```diff
@@ -0,0 +1,90 @@
...
+                interrupt_duration, num_interruptions, delay_between_interruptions, num_active_partitions, interval/1000))
+
+    def drop_traffic_to_port(self, node, port):
+        node.account.ssh_capture("sudo iptables -I INPUT -p tcp --destination-port {} -j DROP".format(port))
+
+    def enable_traffic_to_port(self, node, port):
+        node.account.ssh_capture("sudo iptables -D INPUT -p tcp --destination-port {} -j DROP".format(port))
+
+    def client_server_network_interruption(self, validation_cmd, timeout, interrupt_duration, num_interruptions,
+                                                  delay_between_interruptions, num_active_partitions, processing_duration):
```
zzlbuaa, line 41: super nit: the indentation here doesn't align with the last line.

PR #112, opened 2020-07-30
Ducktape/Vagrant script update

smithakoduri:
The current PR changes handle:
 - Copying respective keystore and trustore files to the respective VMs
 - Create waltz-storage.service and waltz-server.service files
 - Create a common /etc/waltz directory in all VMs to store service specific yml file (for eg: waltz_storage.yml, waltz_server.yml, waltz_client.yml)
 - Any VM can be picked for any service and it will not fail with config missing error any more. (Note: Except Storage service VMs will always have additional disk) 
 - Cleaned up code to be in sync with the current Waltz implementation.

### DIFF
--- a/waltz-test/src/main/python/waltz_ducktape/services/waltz_storage.py
+++ b/waltz-test/src/main/python/waltz_ducktape/services/waltz_storage.py
@@ -5,8 +5,8 @@ class WaltzStorageService(BaseWaltzService):
     """
     WaltzStorageService is the service class for Waltz Storage.
     """
-    def __init__(self, context, cluster_spec, zk, cluster_root, cluster_num_partitions, cluster_key, port, admin_port,
-                 jetty_port, lib_dir, data_dir, config_file_dir):
+    def __init__(self, context, cluster_spec, zk, cluster_root, port, admin_port, jetty_port, lib_dir, data_dir,
+                 config_file_dir, ssl_configs):
         """
         Construct a new 'WaltzStorageService' object.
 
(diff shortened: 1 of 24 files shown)


waltz-test/src/main/python/waltz_ducktape/services/waltz_storage.py
```diff
@@ -5,31 +5,31 @@ class WaltzStorageService(BaseWaltzService):
...
         :param jetty_port: The jetty port
         :param lib_dir: The library directory
         :param data_dir: The data directory
         :param config_file_dir: The directory of configuration file path
+        :param ssl_configs: A dict of ssl related configurations
         """
         super(WaltzStorageService, self).__init__(context, cluster_spec=cluster_spec, zk=zk, cluster_root=cluster_root, port=port, \
                                                   lib_dir=lib_dir, config_file_dir=config_file_dir)
-        self.cluster_num_partitions = cluster_num_partitions
-        self.cluster_key = cluster_key
```
zzlbuaa, line 29: Wonder why do we want to remove the cluster_key here? Is it because now we have `cluster_root` and could fetch those info from it?

smithakoduri: Thats because WaltzStorage fetches this information from the StoreMetadata (via ZooKeeper Connect String). This change was made sometime end of last year. This code was not updated after that.

smithakoduri: I have updated the hardcoded log4j path. Thanks for pointing it out. 👍

PR #118, opened 2020-10-26
Update SSL parsing in the Cli's 

smithakoduri:
Currently, some ServerCli and ClusterCli's are expecting the ssl config to be parsed in "client.ssl" format. Fixed this behavior as these cli's are/mostly executed from the WaltzServer and WaltzStorage nodes and they won't have the ssl config in that format.

After changes, here is the expected ssl parsing format for the CLI's:
 - ServerCli/StorageCli/ClusterCli: "ssl." format
 - ClientCli/PerformanceCli: "client.ssl." format 

Updated Waltz Integration tests to reflect the same.

### DIFF
--- a/waltz-test/src/main/python/waltz_ducktape/tests/validation/load_rebalancing_test.py
+++ b/waltz-test/src/main/python/waltz_ducktape/tests/validation/load_rebalancing_test.py
@@ -18,7 +18,6 @@ class LoadRebalancingTest(ProduceConsumeValidateTest):
 
     def __init__(self, test_context):
         super(LoadRebalancingTest, self).__init__(test_context=test_context)
-        self.server_cli = ServerCli(self.verifiable_client.nodes[0], self.client_config_path)
 
     @cluster(cluster_spec=MIN_CLUSTER_SPEC)
     @parametrize(num_active_partitions=1, txn_per_client=150, num_clients=2, interval=600, timeout=360)
@@ -62,7 +61,8 @@ class LoadRebalancingTest(ProduceConsumeValidateTest):
         server_node = self.waltz_server.nodes[partition_future_server_index]
         server_node_hostname = server_node.account.ssh_hostname
         server = self.get_host(server_node_hostname, server_port)
-        self.server_cli.add_preferred_partition(server, random_active_partition)
+        server_cli = ServerCli(server_node, self.cli_config_path)
+        server_cli.add_preferred_partition(server, random_active_partition)
 
         assert self.get_server_node_idx(random_active_partition) == partition_future_server_index, \
             "partition assignment for server nodes remained untouched"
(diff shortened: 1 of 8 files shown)


waltz-tools/src/test/java/com/wepay/waltz/tools/client/ClientCliTest.java
```diff
@@ -38,12 +39,18 @@ public void restoreStreams() {
         System.setErr(originalErr);
     }
 
-    Properties createProperties(String connectString, String znodePath, int sessionTimeout, SslSetup sslSetup) {
+    Properties createProperties(String connectString,  String znodePath, int sessionTimeout, SslSetup sslSetup) {
+        return createProperties(connectString, znodePath, sessionTimeout, sslSetup,
+            WaltzClientConfig.CLIENT_SSL_CONFIG_PREFIX);
```
patrickghannage, line 44: Nit: not sure if this method is needed. Can't we keep only the generic one below and fill "WaltzClientConfig.CLIENT_SSL_CONFIG_PREFIX" wherever needed ?

waltz-test/src/main/python/waltz_ducktape/tests/validation/load_rebalancing_test.py
```diff
@@ -62,7 +61,8 @@ def rebalancing_on_server_nodes(self, validation_cmd, timeout, num_active_partit
         server_node = self.waltz_server.nodes[partition_future_server_index]
         server_node_hostname = server_node.account.ssh_hostname
         server = self.get_host(server_node_hostname, server_port)
-        self.server_cli.add_preferred_partition(server, random_active_partition)
+        server_cli = ServerCli(server_node, self.cli_config_path)
```
zzlbuaa, line 64: Curious why do we move the ServerCli initialization from the constructor to here and use a different way of fetching server node? Other than this question the PR looks good to me

waltz-tools/src/test/java/com/wepay/waltz/tools/client/ClientCliTest.java
smithakoduri: ohh, I just thought this would be neat compared to adding `WaltzClientConfig.CLIENT_SSL_CONFIG_PREFIX` argument to all createProperties(..) method call.
I can update the code to use the generic createProperties method itself.

waltz-test/src/main/python/waltz_ducktape/tests/validation/load_rebalancing_test.py
smithakoduri: This code change is just to enforce that the serverCli command is executed from one of the waltz-server VM instead of a client VM (as in the constructor we are using verifiable_client.nodes[0]) and with the correct config parser. 

So, that way if executed from the waltz-server VM, it will have the ssl information parsed as expected (i.e. in "ssl." format). Whereas the client VMs have ssl information in their config file parsed in "client.ssl" format.

smithakoduri: We are not modifying the way waltz-server node is fetched. Its still the same (via line 63).

zzlbuaa (approved):

PR #126, opened 2021-01-05
ClusterCli.verify timeout update

smithakoduri:
Updated default timeout to 10sec in ClusterCli.verify cli and created a new optional argument for timeout.


ymatsuda (approved):
Adding the timeout option is a nice idea!

PR #127, opened 2021-01-18
Waltz client metrics implementation

smithakoduri:
Note: Build for this pull request is expected to fail until the dependent change is merged/ released in riff.

### DIFF
--- a/waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
@@ -422,9 +422,9 @@ public class Partition {
 
                 // Send an append message only when the partition is active.
                 if (networkClient != null) {
+                    onCompletionLatencyTimerContext = onCompletionLatencyTimer.time();
                     networkClient.sendMessage(request);
                     sendThroughputMeter.mark();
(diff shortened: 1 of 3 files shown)


smithakoduri:
Have to update the riff version to 2.4.5 once it's released and available in artifactory.

waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
```diff
@@ -683,4 +718,35 @@ public String toString() {
 
     }
 
+    private Boolean isMounted() {
+        synchronized (lock) {
```
ymatsuda, line 722: Is there a reason to use sync here? `mounted` is `volatile boolean`. If you are just observing for metrics, I think sync is unnecessary.

smithakoduri: sorry, my bad. completely missed to notice that the variable is declared as volatile. Will fix it.

ymatsuda (approved):

biplap-sarkar (approved):

git-Bryan (approved):

ymatsuda (approved):

PR #133, opened 2021-05-13
Add/remove preferred partitions to allow batch update

hrdlotom:
Pull request for adding and removing preferred partitions in batches to discuss the initial changes

### DIFF
--- a/waltz-tools/src/main/java/com/wepay/waltz/tools/server/ServerCli.java
+++ b/waltz-tools/src/main/java/com/wepay/waltz/tools/server/ServerCli.java
@@ -190,8 +190,6 @@ public final class ServerCli extends SubcommandCli {
                 }
                 String host = hostAndPortArray[0];
                 int serverPort = Integer.parseInt(hostAndPortArray[1]);
-
-                //int[] partitions = CliUtils.parseIntRanges(partitionId).stream().mapToInt(i -> i).toArray();
                 List<Integer> partitionIds = CliUtils.parseIntRanges(partitionId);
                 addPreferredPartition(host, serverPort, partitionIds, cliConfigPath);
             } catch (Exception e) {
--- a/waltz-tools/src/main/java/com/wepay/waltz/tools/server/ServerCli.java
+++ b/waltz-tools/src/main/java/com/wepay/waltz/tools/server/ServerCli.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Collections;
 
 /**
  * ServerCli is a tool for interacting with Waltz Server.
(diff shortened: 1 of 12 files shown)


waltz-tools/src/main/java/com/wepay/waltz/tools/server/ServerCli.java
```diff
@@ -279,16 +282,18 @@ protected void processCmd(CommandLine cmd) throws SubCommandFailedException {
...
                 throw new SubCommandFailedException(String.format("Failed to remove preferred partition .%n%s", e));
             }
         }
 
-        private void removePreferredPartition(String host, int serverPort, int partitionId, String cliConfigPath) {
+        private void removePreferredPartition(String host, int serverPort, List<Integer> partitionIds, String cliConfigPath) {
+            if (partitionIds.size() == 0) {
+                throw new IllegalArgumentException("Partitions array is empty");
+            }
+
```
smithakoduri, line 296: Same as the above.
Can we also validate the min and max partition id from the given list and verify it is as expected please. Because, we should fail immediately if the partitionId is invalid rather than passing an invalid partition to the server.

waltz-tools/src/test/java/com/wepay/waltz/tools/server/ServerCliTest.java
```diff
@@ -99,7 +102,8 @@ public void testAddAndRemovePreferredPartition() throws Exception {
         Properties cfgProperties = createProperties(helper.getZkConnectString(), helper.getZnodePath(), helper.getSslSetup());
         String configFilePath = IntegrationTestHelper.createYamlConfigFile(DIR_NAME, CONFIG_FILE_NAME, cfgProperties);
 
-        int partitionId = 3;
+        String partitionIdInterval = "3,6-7";
```
smithakoduri, line 105: may be `partitionIdRange` instead of `partitionIdInterval` so that it will be easy to understand.

waltz-tools/src/test/java/com/wepay/waltz/tools/server/ServerCliTest.java
```diff
@@ -116,28 +120,28 @@ public void testAddAndRemovePreferredPartition() throws Exception {
...
             };
             ServerCli.testMain(args1);
 
-            // Verify preferred partition P3 is added to the server descriptor of server 1
+            // Verify preferred partitions P3,P6,P7 are added to the server descriptor of server 1
             serverDescriptors = helper.getWaltzServerRunner().getServerDescriptors();
             for (ServerDescriptor serverDescriptor : serverDescriptors) {
                 if (serverDescriptor.endpoint.toString().equals(serverEndpoint1)) {
-                    assertTrue(serverDescriptor.partitions.contains(partitionId));
+                    partitionIds.forEach(partitionId -> assertTrue(serverDescriptor.partitions.contains(partitionId)));
```
smithakoduri, line 132: Instead of checking for one partition at a time, you can just do `assertTrue(serverDescriptor.partitions.containsAll(partitionIds))`

waltz-tools/src/test/java/com/wepay/waltz/tools/server/ServerCliTest.java
```diff
@@ -116,28 +120,28 @@ public void testAddAndRemovePreferredPartition() throws Exception {
...
 
-                    // Verify partition P3 is assigned to server 1
-                    Boolean partitionMatch = false;
+                    // Verify partitions P3,P6,P7 are assigned to server 1
+                    int partitionMatchIterator = 0;
```
smithakoduri, line 135: partitionMatchCounter instead of partitionMatchIterator may be.

smithakoduri (approved):
Just one small nit. Rest all LGTM.

PR #135, opened 2021-06-17
Waltz client stuck while mounting

hrdlotom:
If waltz client's partition HighWatermark is ahead of server's partition HighWatermark, server notifies client and client's mounting thread is released exceptionally (IllegalStateException is thrown). No resubmission.

Exception handled by individual waltz clients. Treated as any other exception that comes out of submitting a log to waltz.
public void onException(final Throwable throwable) {... is triggered.

### DIFF
--- a/waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
@@ -252,7 +252,7 @@ public class Partition {
      */
     public void partitionAhead() {
         synchronized (lock) {
-            logger.error(String.format("Mounting for partition %d failed, client's high watermark is ahead of server"), partitionId);
+            logger.error(String.format("Mounting for partition %d failed, client's high watermark is ahead of server", partitionId));
             this.clientHighWaterMarkAhead = true;
             lock.notifyAll();
         }
--- a/waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.PriorityQueue;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
@@ -267,6 +268,13 @@ public class Partition {
     public void ensureMounted() {
         if (!mounted) {
             synchronized (lock) {
+                try {
+                    if (clientHWMAhead && this.clientHighWaterMark() <= Long.max(-1L, this.getHighWaterMark().get())) {
+                        networkClient.mountPartition(this);
+                    }
+                } catch (ExecutionException | InterruptedException e) {
+                    logger.error("failed to get high watermark", e.getCause());
+                }
                 while (state != PartitionState.CLOSED && !mounted) {
                     if (transactionMonitor.isStopped()) {
                         throw new PartitionInactiveException(partitionId);
--- a/waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
@@ -7,7 +7,6 @@ import com.wepay.riff.metrics.core.MetricRegistry;
 import com.wepay.riff.metrics.core.Timer;
 import com.wepay.riff.util.Logging;
 import com.wepay.waltz.client.TransactionContext;
-import com.wepay.waltz.client.WaltzClientCallbacks;
 import com.wepay.waltz.client.internal.network.WaltzNetworkClient;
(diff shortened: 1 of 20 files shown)


waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
```diff
@@ -236,24 +239,39 @@ public void unmounted(WaltzNetworkClient networkClient) {
...
         }
     }
 
+    /**
+     * Invoked after receiving MOUNT_RESPONSE informing that local partition' HWM is ahead of server's HWM.
+     * If thread is stuck in ensureMounted, this thread is woken up and throws IllegalStateException
+     */
+    public void partitionAhead() {
+        synchronized (lock) {
+            this.clientHWMAhead = true;
```
smithakoduri, line 255: Let's assume that the client's HWM is ahead of server and partitionAhead() is triggered. This will set the clientHWMAhead as True. And if the client the thread is stuck in "ensureMounted()", it will throw an exception.

How are making sure that eventually (after updating the HWM table) the client will  be able to mount this partition successfully ?

Lets assume there is a transaction written to this partition so "WaltzClient.submit(..)" will be called and eventually "ensureMounted()" will be triggered. But the ensureMounted() will throw an exception again as "clientHWMAhead is set to True". And also note that the client didn't send any MOUNT_REQUEST message again to server in order to be able to mount the partition successfully.

waltz-common/src/main/java/com/wepay/waltz/common/message/MessageCodecV2.java
```diff
@@ -47,8 +47,8 @@ public Message decode(MessageAttributeReader reader) {
                 return new MountRequest(reqId, clientHighWaterMark, seqNum);
 
             case MessageType.MOUNT_RESPONSE:
-                boolean partitionReady = reader.readBoolean();
-                return new MountResponse(reqId, partitionReady);
+                int partitionState = reader.readInt();
```
smithakoduri, line 50: I think we have to update the MessageCodec version as we are changing the partitionState from Boolean to an Integer which means this wont be backward compatible.

smithakoduri (commented):
The changes lgtm.
There are still two fixes to be addressed as part of this pull request:
1. Rename the variable ClientHWMAhead to ClientHighWaterMarkAhead.
2. (Still thinking..little unsure on whether we need this or not) i.e. Update the MessageCodec version as we have added new PartitionStates to the MountResponse message.

waltz-common/src/main/java/com/wepay/waltz/common/message/MessageCodecV0.java
```diff
@@ -120,7 +120,7 @@ public void encode(Message msg, MessageAttributeWriter writer) {
 
             case MessageType.MOUNT_RESPONSE:
                 MountResponse mountResponse = (MountResponse) msg;
-                writer.writeBoolean(mountResponse.partitionReady);
+                writer.writeInt(mountResponse.partitionState);
```
ymatsuda, line 123: This breaks compatibility. Do like this instead.
```
writer.writeBoolean(mountResponse.partitionState == PartitionState.READY);
```

waltz-common/src/main/java/com/wepay/waltz/common/message/MessageCodecV1.java
```diff
@@ -45,8 +45,8 @@ public Message decode(MessageAttributeReader reader) {
                 return new MountRequest(reqId, clientHighWaterMark, seqNum);
 
             case MessageType.MOUNT_RESPONSE:
-                boolean partitionReady = reader.readBoolean();
-                return new MountResponse(reqId, partitionReady);
+                int partitionState = reader.readInt();
```
ymatsuda, line 48: same here

waltz-common/src/main/java/com/wepay/waltz/common/message/MessageCodecV1.java
```diff
@@ -135,7 +135,7 @@ public void encode(Message msg, MessageAttributeWriter writer) {
 
             case MessageType.MOUNT_RESPONSE:
                 MountResponse mountResponse = (MountResponse) msg;
-                writer.writeBoolean(mountResponse.partitionReady);
+                writer.writeInt(mountResponse.partitionState);
```
ymatsuda, line 138: same here

waltz-common/src/main/java/com/wepay/waltz/common/message/MessageCodecV2.java
ymatsuda, line 50: same here

ymatsuda: Yes. We need to create V3 to change the wire protocol.

waltz-common/src/main/java/com/wepay/waltz/common/message/MessageCodecV0.java
hrdlotom: Perhaps if partitionState = CLIENT_AHEAD, we shouldn't send anything. Currently the message will be decoded as NOT_READY leading to more and more mounting calls from the client

ymatsuda (approved):
LGTM

smithakoduri (approved):

PR #138, opened 2021-06-25
GitHub-Actions-CI-Workflow

hrdlotom:
We are looking into replacing Travis-CI workflow with Github Actions


.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Java CI with Gradle
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
biplap-sarkar, line 11: Can we use an LTS version instead of the latest? The idea behind that would be to avoid surprises in the latest version and rely on a version that has long-term support enabled.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Set up JDK 8
+      uses: actions/setup-java@v2
+      with:
+        java-version: '8'
+        distribution: 'adopt'
+    - name: Restore Gradle caches
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.gradle/caches
```
biplap-sarkar, line 24: Can we use paths relative to $HOME? (as done in the deleted .travis.yml file)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+          ~/.gradle/wrapper
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+    - name: Build with Gradle
+      run: ./gradlew build
```
biplap-sarkar, line 32: Same comment for path as in line 24

biplap-sarkar (approved):

PR #150, opened 2021-10-18
Fix to populate current high-water-mark of a partition to commitHighWaterMark variable in AppendTask.init()

smithakoduri:
Currently, the high-water-mark metric in waltz-server/../Partition.java returns the value of "commitHighWaterMark" which is set to Long.MIN_VALUE when the Partition object is created. And, this variable is updated to reflect the latest high-water-mark of a partition only when there is a write to it.

Because of which when a new WaltzServer is created or existing WaltzServer is restarted or when partitions are reassigned to a different server, the high-water-mark metric displays the value as Long.MIN_VALUE for partitions which had data written to it earlier. This ticket will help in resolving that issue and populate the latest high-water-mark of a partition even after the WaltzServer restart.


hrdlotom:
LGTM

hrdlotom (approved):

PR #153, opened 2021-12-06
Server partition mounting logger statements

hrdlotom:
Added extra logger statements to Server Partition. If mounting fails, we should be able to tell why.

### DIFF
--- a/waltz-server/src/main/java/com/wepay/waltz/server/internal/Partition.java
+++ b/waltz-server/src/main/java/com/wepay/waltz/server/internal/Partition.java
@@ -441,8 +441,9 @@ public class Partition {
             if (fetchSize == 0) {
                 // The client is up to date. Send the response immediately. No need to enqueue a feed context.
                 if (!client.sendMessage(response, true)) {
-                    logger.info(String.format("Unable to send mount response message. Channel ctx is null. ClientId: %s, SeqNum: %d, RequestId: %s",
-                        client.clientId(), request.seqNum, request.reqId));
+                    logger.info(String.format("Unable to send mount response message. Channel ctx is null. "
+                            + "ClientId: %s, SeqNum: %d, PartitionId: %d, RequestId: %s",
+                        client.clientId(), request.seqNum, partitionId, request.reqId));
                 }
             } else {
                 // Must bring the client up to date before mounting the partition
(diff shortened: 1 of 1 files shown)


waltz-server/src/main/java/com/wepay/waltz/server/internal/Partition.java
```diff
@@ -407,7 +407,18 @@ public void resumePausedFeedContexts() throws StoreException {
     private boolean isValid(PartitionClient client) {
         synchronized (partitionClientSeqNums) {
             Long currentSeqNum = partitionClientSeqNums.get(client.clientId());
-            return currentSeqNum != null && currentSeqNum.equals(client.seqNum());
+            if (currentSeqNum == null) {
+                logger.info(String.format("CurrentSeqNum is null. Message not valid. ClientId: %s, PartitionId: %d",
```
smithakoduri, line 411: I would rather say "PartitionClient is not valid" instead of "Message not valid".
Lets also print seqNum of the received PartitionClient.

waltz-server/src/main/java/com/wepay/waltz/server/internal/Partition.java
```diff
@@ -428,7 +439,10 @@ private void initializeFeed(MountRequest request, PartitionClient client) {
 
             if (fetchSize == 0) {
                 // The client is up to date. Send the response immediately. No need to enqueue a feed context.
-                client.sendMessage(response, true);
+                if (!client.sendMessage(response, true)) {
+                    logger.info(String.format("Unable to send mount response message. ClientId: %s, SeqNum: %d, RequestId: %s",
```
smithakoduri, line 443: I think it would be better if we mention clearly that the message is not able to send as the Channel ctx is null.
Just a suggestion. That way it will be easier for someone looking at the log statement to understand clearly.

waltz-server/src/main/java/com/wepay/waltz/server/internal/Partition.java
```diff
@@ -428,7 +440,10 @@ private void initializeFeed(MountRequest request, PartitionClient client) {
 
             if (fetchSize == 0) {
                 // The client is up to date. Send the response immediately. No need to enqueue a feed context.
-                client.sendMessage(response, true);
+                if (!client.sendMessage(response, true)) {
+                    logger.info(String.format("Unable to send mount response message. Channel ctx is null. ClientId: %s, SeqNum: %d, RequestId: %s",
```
smithakoduri, line 444: Can we also mention partitionId here.

smithakoduri (commented):
One small comment. Rest LGTM

PR #154, opened 2021-12-06
Show partitions that weren't removed by WaltzServerHandler.

hrdlotom:
Logger. Show partitions that weren't removed by WaltzServerHandler.


waltz-server/src/main/java/com/wepay/waltz/server/internal/WaltzServerHandler.java
```diff
@@ -235,8 +235,15 @@ public void onChannelActive() {
         @Override
         public void onChannelInactive() {
             synchronized (partitions) {
+                List<Integer> partitionsNotRemoved = new ArrayList<>();
```
smithakoduri, line 238: Sorry for nit picking..
Better to use `partitionClientsNotRemoved` instead of `partitionsNotRemoved`.

smithakoduri (approved):

PR #160, opened 2022-03-08
ClusterCli partition generation consistency validation

hrdlotom:
Partition generation on a server may be inconsistent with partition generation metadata in zookeeper store node. Such situation happens for instance if waltz cluster is deleted while server is running and then the cluster is created again with a new storage node. As a result during new session creation waltz server will populate old generation value to Zookeeper and creates a generation inconsistency that will prevent waltz server from mounting this partition after server restarts.

Purpose of this PR is to create a dedicate check for this and give SREs better context of waltz cluster state when running ClusterCli verify.

Steps to locally verify functionality of this PR:

1. Start a new cluster with multiple partitions (delete old waltz containers if needed): 
`export WALTZ_TEST_CLUSTER_NUM_PARTITIONS=3`
`bin/test-cluster.sh start`
2. Stop and start server node couple of times (to increase generation value)
3. Delete waltz cluster `bin/zookeeper-cli.sh delete -n waltz_cluster  --cli-config-path ./config/local-docker/waltz-tools.yml `
4. Create waltz cluster `bin/zookeeper-cli.sh create -p 3 -n waltz_cluster  --cli-config-path ./config/local-docker/waltz-tools.yml`
5. Stop and delete waltz store node docker container
6. Start and populate new store node `bin/test-cluster.sh start`
7. It might be necessary to replace `serverDescriptor.endpoint` in `ClusterCli.buildServersZKPartitionGenerationValidation` with `new Endpoint("localhost", serverDescriptor.endpoint.port)` when running locally
8. Run ClusterCli verify `bin/cluster-cli.sh verify --cli-config-path ./config/local-docker/waltz-tools.yml`

Alongside some other connectivity errors (as ClusterCli verify is not meant to be executed on a localhost), following error message should be visible:
Screen Shot 2022-03-08 at 11 30 03 AM

### DIFF
--- a/waltz-client/src/main/java/com/wepay/waltz/client/internal/InternalRpcClient.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/internal/InternalRpcClient.java
@@ -3,6 +3,7 @@ package com.wepay.waltz.client.internal;
 import com.wepay.waltz.client.WaltzClientCallbacks;
 import com.wepay.waltz.client.internal.network.WaltzNetworkClient;
 import com.wepay.zktools.clustermgr.Endpoint;
+import com.wepay.zktools.clustermgr.PartitionInfo;
 import io.netty.handler.ssl.SslContext;
 
 import java.util.HashMap;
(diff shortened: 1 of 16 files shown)


hrdlotom:
Closed in favor of more general PR: Added server partitions health validation check as part of ClusterCli.Verify - https://github.com/wepay/waltz/pull/168

PR #161, opened 2022-03-17
Specify $platform parameter to support running zookeeper with Apple M1

Hans3q:
the problem:
zookeeper 3.4 docker image does not support Apple M1 and triggers error `no matching manifest for linux/arm64/v8 in the manifest list entries`
the fix:
add parameter `platform=linux/amd64` to `docker run` command


smithakoduri (approved):

hrdlotom (approved):
lgtm

PR #167, opened 2022-04-14
Client connection ordered by generation & seq number

hrdlotom:
**Aim to fix:** We observed time related error when Waltz server is pushed from a cluster and rejoins quickly cluster again once connection to ZooKeeper is resumed. If partition mounting happens with the same seqNum number as was before the connection interruption, it is possible that WaltzServerHandler.OnChannelInactive is called after remounting happened and record of the newly registered NetworkClient is erased by the old NetworkClient.

**Description:** This PR aims to set order to NetworkClients that register clients to server Partitions. This means a NetworkClient with a higher order (N1) shouldn't be replaceable by NetworkClient with equal or lower order (N2) 
`N1.generation > N2.generation or (N1.generation == N2.generation and N1.seqNum >= N2.seqNum)`

**Testing:** New test case is added to verify partitions are successfully mounted even if old NetworkClient record haven't been cleared yet.

Further tests that involves connection shutdown:
Screen Shot 2022-04-14 at 11 56 49 AM

Screen Shot 2022-04-13 at 9 42 39 PM

### DIFF
--- a/waltz-server/src/main/java/com/wepay/waltz/server/internal/Partition.java
+++ b/waltz-server/src/main/java/com/wepay/waltz/server/internal/Partition.java
@@ -68,7 +68,7 @@ public class Partition {
     private final LinkedList<FeedContext> pausedFeedContexts;
     private final FeedCachePartition feedCachePartition;
     private final FeedSynchronizer feedSync = new FeedSynchronizer();
-    private final HashMap<Integer, PartitionClientConnectionOrder> partitionClientConnectionsOrder = new HashMap<>();
+    private final HashMap<Integer, ClientConnectionInfo> partitionClientConnectionInfos = new HashMap<>();
     private final TransactionFetcher transactionFetcher;
 
     private final AtomicBoolean running = new AtomicBoolean(true);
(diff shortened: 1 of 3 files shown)


waltz-server/src/main/java/com/wepay/waltz/server/internal/Partition.java
```diff
@@ -267,14 +267,17 @@ public void receiveMessage(Message msg, PartitionClient client) throws Partition
...
+     * server communicates with client on the latest channel.
      * @param client The client that is interested in this partition.
+     * @param generation The generation of this partition that client is updated to.
      */
-    public void setPartitionClient(PartitionClient client) {
-        synchronized (partitionClientSeqNums) {
-            Long currentSeqNum = partitionClientSeqNums.get(client.clientId());
-            if (currentSeqNum == null || currentSeqNum < client.seqNum()) {
-                partitionClientSeqNums.put(client.clientId(), client.seqNum());
+    public void registerLatestPartitionClient(PartitionClient client, int generation) {
```
smithakoduri, line 275: Why are we renaming the method?

hrdlotom: renamed back. Originally felt like setter might be a little deceiving as passed partitionClient might not get set.

smithakoduri (approved):

PR #174, opened 2022-10-24
EnsureMounted Timeout

hrdlotom:
Introducing timeout of 10s to Client's Partition EnsureMounted(), after which a WaitForMountedTimeoutException is thrown.

This exception acts as other exceptions thrown in this method and is handled by methods invoking waltzClient.submit() that are part of Waltz client services.

Testing: The easiest way to test the code is through WaltzServerTest.testConcurrencyControlMultipleLocks

### DIFF
--- a/waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
+++ b/waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
@@ -17,6 +17,7 @@ import com.wepay.waltz.common.util.Utils;
 import com.wepay.waltz.exception.ClientClosedException;
 import com.wepay.waltz.exception.DataChecksumException;
 import com.wepay.waltz.exception.PartitionInactiveException;
+import com.wepay.waltz.exception.WaitForMountedTimeoutException;
 import com.wepay.zktools.clustermgr.Endpoint;
 import org.slf4j.Logger;
 
@@ -39,6 +40,7 @@ public class Partition {
 
     private static final Long[] EMPTY_LONG_ARRAY = new Long[0];
     private static final int MAX_DATA_ATTEMPTS = 5;
+    private static final int ENSURE_MOUNTED_TIMEOUT = 10000;
 
     private enum PartitionState {
         ACTIVE, INACTIVE, CLOSED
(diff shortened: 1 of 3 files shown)


waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
```diff
@@ -269,17 +271,22 @@ public void partitionAhead() {
...
                     }
-
-                    try {
-                        lock.wait();
-                    } catch (InterruptedException ex) {
-                        Thread.interrupted();
+                    long remaining = due - System.currentTimeMillis();
+                    if (remaining > 0) {
+                        try {
+                            lock.wait(ENSURE_MOUNTED_TIMEOUT);
```
patrickghannage, line 284: Here, do we want to wait on for the remaining amount of time left ?

waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
```diff
@@ -269,17 +271,22 @@ public void partitionAhead() {
...
+                        } catch (InterruptedException ex) {
+                            Thread.interrupted();
+                        }
+                    } else {
+                        throw new WaitForMountedTimeoutException(partitionId, ENSURE_MOUNTED_TIMEOUT);
```
patrickghannage, line 289: Optional suggestion: passing a variable for logging only may not be best practice (even if I see it's been done in Waltz). The reason is it may implicitly suggests that the variable is needed by the class itself for some additional processing. Passing a log message to the Exception class may look cleaner

waltz-client/src/main/java/com/wepay/waltz/client/internal/Partition.java
patrickghannage: hmm, I am curious, there is no unit test for this public method ?

hrdlotom: No, the method doesn't have its own unit test. The method is part of a test flow of other tests such these in InternalStreamClientTest.

patrickghannage (approved):
Looks good to me

PR #178, opened 2023-03-29
feedTask not getting garbage collected

patrickghannage:
When no append requests are happening on a certain partition, feed contexts may accumulate on the real time feed task queue of that partition server in case of a disconnection between a Waltz client and the server holding that partition.

### DIFF
--- a/waltz-server/src/main/java/com/wepay/waltz/server/internal/FeedContext.java
+++ b/waltz-server/src/main/java/com/wepay/waltz/server/internal/FeedContext.java
@@ -12,7 +12,15 @@ import java.util.Comparator;
 public class FeedContext {
 
     public static final Comparator<FeedContext> HIGH_WATER_MARK_COMPARATOR =
-        (o1, o2) -> o1.highWaterMark < o2.highWaterMark ? -1 : o1.highWaterMark > o2.highWaterMark ? 1 : 0;
+        (o1, o2) -> {
+            int hwmCompare = Long.compare(o1.highWaterMark, o2.highWaterMark);
+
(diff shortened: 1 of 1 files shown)


waltz-server/src/main/java/com/wepay/waltz/server/internal/FeedContext.java
```diff
@@ -12,7 +12,15 @@
 public class FeedContext {
 
     public static final Comparator<FeedContext> HIGH_WATER_MARK_COMPARATOR =
-        (o1, o2) -> o1.highWaterMark < o2.highWaterMark ? -1 : o1.highWaterMark > o2.highWaterMark ? 1 : 0;
+        (o1, o2) -> {
+            int hwmCompare = Long.compare(o1.highWaterMark, o2.highWaterMark);
+
+            if (hwmCompare == 0) {
+                return Integer.compare(o1.reqId.seqNum(), o2.reqId.seqNum());
```
hrdlotom, line 19: Let's say we have FeedContexts of equal HWM. The first feed context is still going to block the queue while waiting for a new transaction. At this point the FeedContext becomes inactive but is still blocking the queue that may be dangerously grow. This doesn't critically influence the functionality of Waltz as with new transaction old expired FeedContexts are all erased and active FeedContexts are served. 

```
Partition.FeedTask.process:

if (highWaterMark < nextTransactionId) {
     // Wait for more data or new feed context
    feedSync.await(version);
}```

hrdlotom (approved):
LGTM. Expired feed contexts will be poped out when new feed context is inserted.

